PR #5, opened 2014-06-26
Add a filter on mako entities to ensure XML entities are converted.

nbessi:
Fix lp: 132415


nbessi:
Setting default_filter to XML to avoid error when validating the XML file.

http://docs.makotemplates.org/en/latest/filtering.html#the-default-filters-argument

bwrsandman:
Please rebase.

nbessi:
Correct target branch is 7.0

PR #25, opened 2014-07-02
Add coverage config file and README badges

bwrsandman:
To celebrate tests passing, why not add the status badges!


nbessi:
Thanks for the addition.

Who only 42 percent of coverage there is somme work to do on this topic...

PR #26, opened 2014-07-03
Update .travis.yml

nbessi:


bwrsandman:
The command should be 

`export PATH=...`

bwrsandman:
> The command should be 

Sorry, I thought it had failed, looks like it works both ways.

bwrsandman:
Sorry for fb66b0a8d3, I meant for it to be a PR, but I guess having write access makes it so I edit it directly when using the web editor.

Anyway, fb66b0a8d3 is a minor change to the readme badge and shouldn't cause you any grief.

Still, I'll make sure I don't commit directly again.

PR #33, opened 2014-07-31
update travis config

yvaucher:
Update travis with last changes on maintainer-quality-tools

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,26 +5,24 @@ language: python
 python:
   - "2.7"
 
+env:
+  - VERSION="7.0" ODOO_REPO="odoo/odoo"
+  - VERSION="7.0" ODOO_REPO="OCA/OCB"
+
 virtualenv:
   system_site_packages: true
 
-before_install:
-  - git clone https://github.com/OCA/partner-contact-management $HOME/partner-contact-management -b 7.0
-  - git clone https://github.com/OCA/bank-statement-reconcile $HOME/bank-statement-reconcile -b 7.0
-  - git clone https://github.com/OCA/account-financial-tools $HOME/account-financial-tools -b 7.0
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -5,26 +5,24 @@ language: python
...
+  - git clone https://github.com/OCA/account-financial-tools {$HOME}/account-financial-tools -b$ {VERSION}
+  - git clone https://github.com/OCA/maintainer-quality-tools.git {$HOME}/maintainer-quality-tools
+  - export PATH={$HOME}/maintainer-quality-tools/travis:$PATH
+  - travis_install_nightly ${VERSION}
 
 script:
-    # - travis_run_flake8 will be activaed in a separated commit
-    - travis_run_tests 7.0 $HOME/partner-contact-management $HOME/bank-statement-reconcile $HOME/account-financial-tools
+    - travis_run_flake8
+    - travis_run_tests ${VERSION}
```
pedrobaeza: I think this is still not auto-detected, so you have to put explicitly the dependencies list

yvaucher: According to maintainer-quality-tools README:

"""The addons path used will automatically consider these repositories."""

But this is my first try. No idea if it will work.

pedrobaeza: Maybe an incorrect README? What we were talking about is to put a list of repositories to be downloaded to an specific folder and added to the addons-path automatically, but you're downloading dependencies manually.

Let's see what Travis says :wink:

yvaucher: Seams to work

https://github.com/OCA/maintainer-quality-tools/blob/master/travis/travis_run_tests

The magic is there
https://github.com/OCA/maintainer-quality-tools/commit/c89cbdfa1ef73de9741e00413f07c0978cdeedc9

pedrobaeza:
:+1:

PR #50, opened 2014-09-23
Fixes Errors when changing company's bank informations

renzo1974:
found in 8.0 stable from 20140922


nbessi:
Hello,
Thanks for your patch
It is strange to do a patch on an **unported** except to rebase with 7. 0.

The port of localisation will change the structure of the code to integrate it with the new API et code
base. So only at taht time only fully ported module should be proposed to merge.
At that time  you should not  install unported addons on your production server.

There is some active work on the localistion port here:
https://github.com/PCatinean/l10n-switzerland

The patch is also weird as we are using a browse record the id should not be passed to function.
It is maybe a bug in upstream.

PR #54, opened 2014-10-29
Add commercial_partner_id on supplier invoice and

vrenaville:
set it with on_change_partner_id in order to have
company bank account in the related field selection

### DIFF
--- a/l10n_ch_base_bank/invoice.py
+++ b/l10n_ch_base_bank/invoice.py
@@ -52,7 +52,8 @@ class AccountInvoice(Model):
                     to be able to filter in the view in the field
                     partner_bank_id
                     """
-                    res['value']['commercial_partner_id'] = p.commercial_partner_id.id
+                    p_id = p.commercial_partner_id.id
+                    res['value']['commercial_partner_id'] = p_id
                 if p.bank_ids:
                     bank_id = p.bank_ids[0].id
                 elif p.commercial_partner_id.bank_ids:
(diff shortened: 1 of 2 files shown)


vrenaville:
@nbessi Could you please review my change

l10n_ch_base_bank/invoice.py
```diff
@@ -46,8 +46,21 @@ def onchange_partner_id(self, cursor, uid, ids, invoice_type, partner_id,
                     partner_id,
                     context
                 )
+                if p.commercial_partner_id:
+                    """
+                    We set the commercial_partner_id in the view
+                    to be able to filter in the view in the field
+                    partner_bank_id
+                    """
+                    res['value']['commercial_partner_id'] = p.commercial_partner_id.id
```
nbessi: Please add an intermediate variable

nbessi:
One pep8 issue but else LGTM. Please can you create a issue for the record

vrenaville:
@nbessi  thanks for the review I have just fixed it

mdietrichc2c:
:+1:

nbessi:
If it is green :+1:

fclementic2c:
Does not work if the supplier invoice is created from a PO.
Otherwise, it is Ok

vrenaville:
@fclementic2c The purchase order doen't call onchange partner, but with my fix the user can select bank account from the commercial_partner_id, so it work, but it's need a manual operation.
It can be an improvement but it's not a side effect of my PR

fclementic2c:
ok then :+1:

PR #61, opened 2014-11-18
Add check for transaction id before calling SQL request

mdietrichc2c:
This PR is to check that the transaction ID is present before calling the SQL request; if the value is false, a SQL error "boolean != text" is returned, and the transaction block is rollbacked for the remainder of the auto-completion.


guewen:
Could you do:

```
if not st_line['transaction_id']:
    return res
```

Instead of pushing the whole block to the right?

nbessi:
I agree with @guewen

mdietrichc2c:
@nbessi  @guewen sorry, missed that PR. I fixed it as @guewen said.

nbessi:
:+1:

yvaucher:
:+1:

PR #75, opened 2014-12-22
l10n_ch_base_bank breaks onchange_partner_id on invoices

guewen:
- Remove the useless view that breaks the onchange
- Use 'attributes' instead of 'replace' and change only the domain
- Remove the deprecated 'type' of views


hurrinico:
:+1:

nbessi:
:+1:

PR #81, opened 2015-01-16
[account_statement_import] Move attachements from statement to statement_line

ecino:
In statement import module, the Postfinance parser was attaching pictures to the bank statement object. Those pictures are however corresponding to a specific statement line.

I propose to move the attachments directly to the statement line objects and added a button in the view in order to be able to display the attachment right from bank statement view.


guewen:
:+1:

nbessi:
:+1:

davidwul:
Nice addition, it works well :+1:

PR #86, opened 2015-01-21
[ADD] [8.0] l10n_ch_payment_slip_voucher

eLBati:
we have some small customers which find vouchers convenient. The same small customers usually apply the VAT on payment regime.
So, we created this module mainly to apply the VAT on payment regime (that at the moment is voucher based)


nbessi:
@fclementic2c  can you please do a business review

l10n_ch_payment_slip_voucher/tests/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+from . import test_v11_import
+fast_suite = [
+    test_v11_import
+]
```
guewen: `fast_suite` is no longer used in v8 and should be removed (tests are loaded as soon as they are imported in this file)

l10n_ch_payment_slip_voucher/tests/test_v11_import.py
```diff
@@ -0,0 +1,83 @@
...
+                                       'test1.v11')
+        with open(v11_path) as v11_file:
+            importer = self.env['v11.import.wizard.voucher'].create({
+                'v11file': base64.encodestring(v11_file.read()),
+                'currency_id': self.env.ref('base.EUR').id,
+                'journal_id': self.env.ref('account.bank_journal').id,
+                })
+            std_importer = self.env['v11.import.wizard'].create({})
+            v11_file.seek(0)
+            lines = v11_file.read().split("\r\n")
```
guewen: Couldn't you use `readline()`?

l10n_ch_payment_slip_voucher/wizard/bvr_import.py
```diff
@@ -0,0 +1,165 @@
...
+                    'res_model': 'account.voucher',
+                    'res_id': voucher_id,
+                },
+            )
+
+        action_res = self.env['ir.actions.act_window'].for_xml_id(
+            'account_voucher', 'action_vendor_receipt')
+        action_res[
+            'domain'
+        ] = "[('id','in', ["+','.join(map(str, voucher_ids))+"])]"
```
guewen: you could simply provide it in python, it works too:

```
action_res['domain'] = [('id' ,'in', voucher_ids)]
```

Or with a string but simpler (`%s` will repr the list the same way):

```
action_res['domain'] = "[('id' ,'in', %s)]" % (voucher_ids,)
```

guewen:
I guess you had had a pleasant time dealing with the voucher's api :grin: 
:+1: (a few minor remarks)

l10n_ch_payment_slip_voucher/tests/test_v11_import.py
eLBati: I followed the same approach used at
https://github.com/OCA/l10n-switzerland/blob/8.0/l10n_ch_payment_slip/wizard/bvr_import.py#L234
for coherence

eLBati:
Thanks @guewen I made some improvements

nbessi:
@fclementic2c any news on this one

l10n_ch_payment_slip_voucher/tests/test_v11_import.py
guewen: I would have preferred if `readlines()` or `splitlines()` was used, even if it is different in the other one (which should be fixed IMO). Those methods will work regardless of the OS.

nbessi:
:+1:

nbessi:
For me it is fine. As soon @fclementic2c confirms it is correct in term of business we can merge

fclementic2c:
I do not have time to test sorry. My worry is : 
- Does it work in case of multicurrency with a currency difference ?
- Does it work in case of multicurrency and partial payment ?

l10n_ch_payment_slip_voucher/tests/test_v11_import.py
eLBati: Right @guewen : https://github.com/eLBati/l10n-switzerland/commit/43531e500249a062add57710a6d8ffaf7766331c

eLBati:
Hello @fclementic2c , currency is selected in wizard and set in voucher as is.
Then, to handle currency differences and partial payments in case of multi currency is a voucher task.
Do you see problems with that?

nbessi:
Ok seen with @fclementic2c

PR #92, opened 2015-02-04
[FIX] wrong scan line

eLBati:
Without this change, invoice with total amount of 4502.47 CHF produces the line
00014502475>...

With this change, it produces the line
0100004502471>...

### DIFF
--- a/l10n_ch_payment_slip/payment_slip.py
+++ b/l10n_ch_payment_slip/payment_slip.py
@@ -202,8 +202,8 @@ class PaymentSlip(models.Model):
         line = []
         if not self._can_generate(self.move_line_id):
             return []
-        amount = '01%.2f' % self.amount_total
-        justified_amount = amount.replace('.', '').rjust(10, '0')
+        justified_amount = '01%s' % unicode(
+            self.amount_total).replace('.', '').rjust(10, '0')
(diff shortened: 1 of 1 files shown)


nbessi:
After reading https://www.postfinance.ch/binp/postfinance/public/dam.Up1Ct9TDDT8MFk4AB6m00phIHrTLJlvd0XyZjBIXy3M.spool/content/dam/pf/de/doc/consult/manual/dlserv/inpayslip_isr_man_fr.pdf 

fix seems correct let see if test are green.

l10n_ch_payment_slip/payment_slip.py
```diff
@@ -202,8 +202,8 @@ def _compute_scan_line_list(self):
         line = []
         if not self._can_generate(self.move_line_id):
             return []
-        amount = '01%.2f' % self.amount_total
-        justified_amount = amount.replace('.', '').rjust(10, '0')
```
nbessi: We have to round the amount_total to 2 digit to ensure bvr is on 2 digit it precision is changed.

nbessi:
:+1:

nbessi:
@eLBati it would be really nice if you can add a test that check scan line see #93

guewen:
:+1:

PR #103, opened 2015-02-23
[FIX] [8.0]  l10n_ch_payment_slip : line_ids is a recordset

eLBati:
Otherwise, importing V11 file:
`ProgrammingError: can't adapt type 'account.move.line'`

### DIFF
--- a/l10n_ch_payment_slip/wizard/bvr_import.py
+++ b/l10n_ch_payment_slip/wizard/bvr_import.py
@@ -206,7 +206,7 @@ class BvrImporterWizard(models.TransientModel):
         )
         if line_ids:
             # transaction_ref is propagated on all lines
-            line = move_line_obj.browse(line_ids[0])
+            line = line_ids
             partner_id = line.partner_id.id
             num = line.invoice.number if line.invoice else False
             values['ref'] = _('Inv. no %s') % num if num else values['name']
--- a/l10n_ch_payment_slip/wizard/bvr_import.py
+++ b/l10n_ch_payment_slip/wizard/bvr_import.py
@@ -197,16 +197,18 @@ class BvrImporterWizard(models.TransientModel):
                   'type': (record['amount'] >= 0 and 'customer') or 'supplier',
                   'statement_id': statement.id,
                   }
-        line_ids = move_line_obj.search(
+        line = move_line_obj.search(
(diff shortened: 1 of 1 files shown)


l10n_ch_payment_slip/wizard/bvr_import.py
```diff
@@ -206,7 +206,7 @@ def _prepare_line_vals(self, statement, record):
         )
         if line_ids:
```
yvaucher: In fact `line_ids` is not named properly as it is a record set.
Plus can we have multiple lines returned by search?
If yes and we still want to manage only one, we can add `limit=1` on search

Please add `limit=1` and rename `line_ids` to `line` and remove `line = line_ids`

yvaucher:
@eLBati Thanks for the patch. See my in-line comment for some little improvements.

eLBati:
@yvaucher I think we can't have multiple lines returned by search (@nbessi is it true?). So I added the check

l10n_ch_payment_slip/wizard/bvr_import.py
```diff
@@ -197,16 +197,18 @@ def _prepare_line_vals(self, statement, record):
...
             [('transaction_ref', '=', reference),
              ('reconcile_id', '=', False),
              ('account_id.type', 'in', ['receivable', 'payable']),
              ('journal_id.type', '=', 'sale')],
             order='date desc',
         )
-        if line_ids:
+        if len(line) > 1:
+            raise exceptions.Warning(
+                "Too many credit lines for reference %s" % reference)
```
nbessi: We should replace credit with receivable/payable in order not to get confused with the credit line of credit control. The exception message should be translatable.

nbessi:
@eLBati from a business point of  view this case may happens with multi payment term invoices. The proposed message is better than the one raised by ensure_one. 

So :+1:  for the fix once the word `credit` is replaced and the translation file updated

yvaucher:
:+1:

eLBati:
@nbessi thanks, fixed.

> from a business point of view this case may happens with multi payment term invoices

In this case, how could the move.line be identified?

nbessi:
@eLBati you need to use other addons like the one provided in the bank_statement_reconcile project

mdietrichc2c:
:+1: for this fix.

PR #112, opened 2015-03-04
Allow canceling draft invoices

ghost:


yvaucher:
:+1:  It's ok, pep8 error is unrelated

guewen:
Not sure about the correctness of this change. Previously, the check was done when the invoice was all but `draft` and now it is only done when it is `draft` or `cancel`.

yvaucher:
@guewen Oh wait it seems you are right good catch

@BT-csanchez Don't you mean:

```
if invoice.state in ('draft', 'cancel'):
    continue
```

The `not` shouldn't be there I think.

l10n_ch_base_bank/invoice.py
```diff
@@ -103,7 +103,9 @@ def _check_bvr(self, cr, uid, ids, context=None):
         """
         invoices = self.browse(cr, uid, ids)
         for invoice in invoices:
-            if invoice.reference_type == 'bvr' and invoice.state != 'draft':
+            if invoice.state not in ('draft', 'cancel'):
```
nbessi: Condition seems inverted not in/in

ghost:
Hello, 

The idea is to extend the behavior not only for those
 invoices that have their state in draft but in cancel. 

Should I change the code in order to make it more comprehensive?

yvaucher:
@BT-csanchez right we understood that, but condition is inverted. You want to continue if in ('draft', 'cancel') no ?

We want to bypass the check only if `draft` or `cancel`

Why would you do the check if state is `cancel` ?

nbessi:
any answer on this one

ghost:
I'm going to rewrite the condition and commit it again with some comments

ghost:
The underline idea is not check bvr on draft and cancel invoices.
Previously only it was set on draft invoices.

yvaucher:
@BT-csanchez Thanks for the correction

This time the condition is right as not inverted by the `continue` logic

Just few PEP8 issues to fix and we will be good:

```
./l10n_ch_base_bank/invoice.py:107:72: E271 multiple spaces after keyword
./l10n_ch_base_bank/invoice.py:107:80: E501 line too long (92 > 79 characters)
./l10n_ch_base_bank/invoice.py:107:82: E231 missing whitespace after ','
```

yvaucher:
Ok this was fixed, I'll fix the pep8 and merge :+1:

yvaucher:
Merged at https://github.com/OCA/l10n-switzerland/commit/227722f68befcfab915ad41d40160c060cf9e683

yvaucher:
Thanks @BT-csanchez

PR #118, opened 2015-03-27
[FIX] Preserve move transaction_ref, when you cancel and reopen an invoi...

vrenaville:
This fix prevent transaction_ref to have a different number 
when you cancel and reopen an invoice.
The problem occurs if you have already sent to your customer the BVR report, so I will pay you with a wrong reference and matching payment will not been done automatically.
I have use a sequence counter to numerate line for multiple payment term in order to have a different reference for each terms.

@nbessi your review on this one will be highly  appreciate

### DIFF
--- a/l10n_ch_payment_slip/invoice.py
+++ b/l10n_ch_payment_slip/invoice.py
@@ -37,7 +37,8 @@ class AccountMoveLine(Model):
         """Hook to get amount in CHF for BVR"""
         return move.debit
 
-    def get_bvr_ref(self, cr, uid, move_line_id, context=None):
+    def get_bvr_ref(self, cr, uid, move_line_id,
+                    sequence_int=0, context=None):
         """Retrieve ESR/BVR reference from move line in order to print it
 
         Returns False when no BVR reference should be generated.  No
@@ -52,13 +53,11 @@ class AccountMoveLine(Model):
 
         if not self._is_generate_bvr(cr, uid, move_line.invoice, context):
             return ''
-
-        reference = self._compute_bvr_ref(cr, uid, move_line, context)
-        if (move_line.transaction_ref and
-                move_line.transaction_ref != reference):
-            # the line has already a transaction id and it is not
-            # a BVR reference
-            return ''
+        if move_line.transaction_ref:
+            return move_line.transaction_ref
+        reference = self._compute_bvr_ref(cr, uid,
+                                          move_line,
+                                          sequence_int, context)
         return reference
 
     def _is_generate_bvr(self, cr, uid, invoice, context=None):
@@ -67,15 +66,15 @@ class AccountMoveLine(Model):
         return (invoice.partner_bank_id and
                 invoice.partner_bank_id.state == 'bvr')
 
-    def _compute_bvr_ref(self, cr, uid, move_line, context=None):
+    def _compute_bvr_ref(self, cr, uid, move_line,
+                         sequence_int=0, context=None):
         ''' Default BVR reference computation '''
         res = ''
         if move_line.invoice.partner_bank_id.bvr_adherent_num:
             res = move_line.invoice.partner_bank_id.bvr_adherent_num
-        move_number = ''
-
+        move_number = str(sequence_int)
         if move_line.invoice.number:
-            compound = str(move_line.invoice.number) + str(move_line.id)
+            compound = str(move_line.invoice.number) + str(sequence_int)
             move_number = self._compile_get_ref.sub('', compound)
         reference = mod10r(res + move_number.rjust(26 - len(res), '0'))
         return reference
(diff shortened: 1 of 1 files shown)


vrenaville:
@CSester Could you please review my PR, I have revert a part of your code, to be able to print consistent BVR reference in BVR report.

rdeheele:
LGTM :+1:

l10n_ch_payment_slip/invoice.py
```diff
@@ -102,26 +101,24 @@ def _compute_full_bvr_name(self, cr, uid, ids, field_names, arg,
...
-            [('type', 'in', ['receivable', 'payable'])],
-            context=context
-        )
         for inv in self.browse(cr, uid, ids, context=context):
             move_lines = move_line_obj.search(
                 cr,
                 uid,
                 [('move_id', '=', inv.move_id.id),
-                 ('account_id', 'in', tier_account_id)],
+                 ('account_id', '=', inv.account_id.id)],
```
rdeheele, line 109: are we sure that all move lines that need a bvr_ref have always the same account_id than the invoice?

l10n_ch_payment_slip/invoice.py
```diff
@@ -202,15 +199,19 @@ def action_number(self, cr, uid, ids, context=None):
...
                 continue
             move_lines = move_line_obj.browse(cr, uid, move_line_ids,
                                               context=context)
+            sequence_int = 0
             for move_line in move_lines:
+                sequence_int += 1
                 if inv.type in ('out_invoice', 'out_refund'):
-                    ref = move_line.get_bvr_ref()
+                    ref = move_line.get_bvr_ref(sequence_int=sequence_int,
+                                                context=context)
```
guewen, line 214: the context should never be passed manually when a method is called on a browse_record, it is already contained in the record when `browse()` is called.

l10n_ch_payment_slip/invoice.py
```diff
@@ -67,15 +66,15 @@ def _is_generate_bvr(self, cr, uid, invoice, context=None):
...
         ''' Default BVR reference computation '''
         res = ''
         if move_line.invoice.partner_bank_id.bvr_adherent_num:
             res = move_line.invoice.partner_bank_id.bvr_adherent_num
-        move_number = ''
-
+        move_number = str(sequence_int)
         if move_line.invoice.number:
-            compound = str(move_line.invoice.number) + str(move_line.id)
+            compound = str(move_line.invoice.number) + str(sequence_int)
```
guewen, line 77: Can you explain why it is better to use this `sequence_int` instead of `move_line.id`, in a comment maybe? I don't understand it.

l10n_ch_payment_slip/invoice.py
guewen, line 109: I think that's correct

l10n_ch_payment_slip/invoice.py
guewen, line 77: My feeling is that it makes the bvr ref computation very fragile because it depends on how the method is called and there is chances that the caller does not pass the correct `sequence_int` to obtain the correct bvr ref.

l10n_ch_payment_slip/invoice.py
```diff
@@ -52,13 +53,11 @@ def get_bvr_ref(self, cr, uid, move_line_id, context=None):
...
             return ''
-
-        reference = self._compute_bvr_ref(cr, uid, move_line, context)
-        if (move_line.transaction_ref and
-                move_line.transaction_ref != reference):
-            # the line has already a transaction id and it is not
-            # a BVR reference
-            return ''
+        if move_line.transaction_ref:
+            return move_line.transaction_ref
```
guewen, line 57: What is the reason to change this condition?

l10n_ch_payment_slip/invoice.py
vrenaville, line 214: Thanks, I will correct that

l10n_ch_payment_slip/invoice.py
vrenaville, line 77: Yes, It fragile, but we need to have a number that must be identical if you cancel and reconfirmed the invoice.
as soon as the move line is different each time you reopen the invoice, you will get a different number, if you have already send the BVR report to your customer, you will not be able to match the payment with the invoice automatically

l10n_ch_payment_slip/invoice.py
vrenaville, line 57: The BVR report  call the get_bvr_ref without an sequence_int parameter (I cannot determine the sequence in the report), so if the move line have already a transaction_ref I simply return it.
The test at line #L54 already implement a test, if it's a BVR or not, so at #L56 we have a valid transaction_ref.

nbessi:
Travis run failed due to timeout. Run restarted

mdietrichc2c:
Apart from the `context=context` issue described by @guewen, looks good to me. :+1:

vrenaville:
After a second analysis, my PR is too fragile, I will propose another one that allow person to modify unimportant field on the invoice.

PR #121, opened 2015-03-31
Remove wrong usage of @api.v7

guewen:
This decorator must only be used when @api.v8 is used in the same Model, it allows
to dispatch v8 and v7 calls to methods with the same name but with different signatures.

When used alone, if a v8 call is done to the method, it tries to call NoneType.


yvaucher:
:+1:

eLBati:
:+1:

nbessi:
:+1:

PR #123, opened 2015-04-09
[FIX] Prevent traceback when computing a bvr without reference

PCatinean:
When the bvr reference is False the _compute_full_bvr_name in invoice throws an error trying to join the False values


guewen:
:+1:

eLBati:
:+1:

PR #111, opened 2015-03-04
Bugfixing in domain

ghost:

PR #125, opened 2015-04-15
Revert "Bugfixing in domain"

nbessi:
Reverts OCA/l10n-switzerland#111


yvaucher:
@nbessi What about this PR?

PR #138, opened 2015-06-03
bvr_reference importing

vrenaville:
I replace the reference add in the DTA file by the payment.line communication field

this communication field is field with bvr_reference in case of a bvr invoice
and a stand invoice ref in other case, when you import invoice in the payment order


l10n_ch_dta/wizard/create_dta.py
```diff
@@ -607,7 +607,8 @@ def _set_bank_data(self, pline, elec_context, seq):
         elec_context['partner_bank_country'] = b_country
 
         elec_context['partner_bank_code'] = pline.bank_id.bank_bic
-        elec_context['reference'] = pline.move_line_id.ref
+        elec_context['reference'] = pline.move_line_id.transaction_ref \
+            or pline.move_line_id.ref or False
```
nbessi: Please use if else form also no need for the False

nbessi:
Seems good to me just some details to fix. An issue or a full explaination of the bug on the comment would be appreciated. @fclementic2c  does it seems ok in term of business?

nbessi:
Test must be fixed so

fclementic2c:
:+1:

vrenaville:
@nbessi  thanks for the review

This fix add the check of BVR reference on supplier invoice, and add the generation of transaction_ref for supplier invoice for supplier invoice (currently we will give the same transaction_ref to all related account_move_line, in case of multiple payment term).
This transaction_ref (if exist) is import in the payment order in the communication field.
So when you generated the DTA from the payment order, it works as expected as soon as you have a valid transaction_ref in the move line.

I hope it's more clear now.

Vincent

nbessi:
:+1:

PR #141, opened 2015-06-10
Port statement import

nbessi:


nbessi:
@ecino I created the PR here feel free to update it or the close it

davidwul:
@nbessi why attaching the scanned BV to the statement instead of the statement line? Attaching to the statement line would allow to link easily the image to relevant record. With the actual behavior it's pretty hard to identify which image goes with which statement line. The usability is therefore pretty poor. 

Also attaching the xml to the statement could be useful. 
What do you think?

yvaucher:
See travis result, you have got some PEP8 issues to fix

nbessi:
@davidwul With the current implementation of base statement import module  it his hard to directly specify the attachment at line level. In order to do this we have to do a re-factoring of the base module. If you think it is a need that is going to be commonly needed just goes to @fclementic2c and we see what we can do . 

That said we can post process the attachment of the statement to link them to the statement lines as with Postfinance the attachment have a naming convention with the statement line description. Just tell me if you want me to proceed with this task.

Attaching the xml should be quite an easy task and seem a legitimate need.

nbessi:
Travis failed due to issue https://github.com/OCA/bank-statement-import/issues/26

davidwul:
@nbessi I don't think a refactor would be needed because (hopefully) image attachment should disappear with the bank migration toward ISO20022 (camt052 etc..). A post process is a good idea. Having it link to the statement line is not a hard requirement. The ultimate goal would be to be able to display the image with a single click (similar to what we have in v7) from within the (new v8) reconciliation process. 
Could you handle that?

nbessi:
@davidwul I'm not sure to understand the need. You should check with @fclementic2c.

davidwul:
@nbessi To explain the need: here is (below) an excerpt of one of our postfinance bank statements. All lines underlined in red are the ones that have a image. The only text info we have is in the OBI: "BULLETIN DE VERSEMENT" which is not really informative.(as opposed to the lines crossed out in green ) So the need would be in the reconciliation process to have the image displayed (or clickable) so the completion and purpose of the donation can be known easily. If it's still unclear I'll be happy to talk with @fclementic2c 
bs

ecino:
We can close this one, work was continued on #147

PR #148, opened 2015-07-01
[7.0-lsv-dd] fix : mandate_ids list must not contain duplicates

ecino:
Bug fix where a list could contain twice or more the same ids causing the write method to raise an error.


ecino:
Can anyone review this one ? It is a very small PR and a bug fix we could benefit.

guewen:
:+1: 
Thanks for the fix!

yvaucher:
:+1:

PR #153, opened 2015-07-24
add the possibility to validate every created voucher while importing…

eLBati:
… file


nbessi:
:+1: but I realized that po file are missing. @eLBati it would be nice if you can add them in an other PR

eLBati:
Thanks @nbessi here it is https://github.com/OCA/l10n-switzerland/pull/154

nbessi:
Nice @eLBati

yvaucher:
:+1:

PR #159, opened 2015-08-14
[FIX] when bank name is too long, it overlaps with 'giro' bank

eLBati:
See example with 'Banca Raiffeisen del Cassarate':

bvr

### DIFF
--- a/l10n_ch_payment_slip/payment_slip.py
+++ b/l10n_ch_payment_slip/payment_slip.py
@@ -23,6 +23,7 @@ import base64
 import StringIO
 import contextlib
 import re
+import textwrap
 from collections import namedtuple
 from reportlab.pdfgen.canvas import Canvas
 from reportlab.pdfbase import pdfmetrics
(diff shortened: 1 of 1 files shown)


l10n_ch_payment_slip/payment_slip.py
```diff
@@ -547,7 +548,8 @@ def _draw_bank(self, canvas, print_settings, initial_position, font, bank):
         text = canvas.beginText()
         text.setTextOrigin(x, y)
         text.setFont(font.name, font.size)
-        lines = bank.name.split("\n")
+        bank_name = textwrap.fill(bank.name, 26)
```
eLBati, line 551: @nbessi is 26 always valid? Should it be parametric?

yvaucher, line 551: This font seems monospaced. So it should be ok.

yvaucher:
:+1:

nbessi:
:+1:

PR #173, opened 2015-10-14
Remove l10n ch

yvaucher:
`l10n_ch` module is fully maintained in Odoo Community since 7.0 as it only contain Chart of account


tafaRU:
:+1:

vrenaville:
:+1:

PR #178, opened 2015-10-19
[FIX] Do not set a default partner on move line

vrenaville:
It's no useful force a partner on move line


yvaucher:
This module definitely needs some tests, can you add one ?

PR #180, opened 2015-10-22
Remove l10n_ch

yvaucher:
this module is not needed anymore as it is updated in official Odoo code
base, and in case it isn't fixes will land in OCB.

Plus this uninstallable module overwrite the core one making l10n_ch not
working.


yvaucher:
As this PR is for cleaning the code base and make tests works I fast-track it. 
Plus it has already been done for 9.0 branch

PR #184, opened 2015-10-26
[9.0] [port] l10n_ch_bank

yvaucher:
Depends:
- [x] #179 

post_init_hook due to https://github.com/odoo/odoo/issues/10696


philhug:
@yvaucher there seems to be a small bug in the bank.xml
Here's a patch:

diff --git a/l10n_ch_bank/data/bank.xml b/l10n_ch_bank/data/bank.xml
index 29a32fb..61215b1 100644
--- a/l10n_ch_bank/data/bank.xml
+++ b/l10n_ch_bank/data/bank.xml
@@ -86295,7 +86295,7 @@
         <field name="bank_postaccount">*19-6989-5</field>
         <field name="active" eval="True"/>
     </record>
-    <record model="res.bank" id="bank_80539_No entry found for host https://www.transifex.com. Creating...">
-    <record model="res.bank" id="bank_80539_0009">
       <field name="name">Raiffeisenbank Goms</field>
       <field name="clearing">80539</field>
       <field name="code">Raiffeisen</field>

yvaucher:
@philhug thanks I see there is still some issues in ids due to my last commit

yvaucher:
@philhug note that you can format your code in message by putting `back quotes` or

```
adding 4 spaces
at start of line
```

Your patch wasn't readable but I found the issue

This is now fixed

l10n_ch_bank/models/res_config.py
```diff
@@ -1,25 +1,7 @@
...
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
-
+# © 2011-2014 Nicolas Bessi (Camptocamp SA)
+# © 2014 brain-tec AG
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 """
 Adds a button allowing to reload all the banks from the XML file.
 Thus, if this module has already been installed but the data is changed,
```
yvaucher: To fix this feature is not working anymore as loading data has changed

yvaucher:
@BT-ojossen Hello could you please review this? I replaced the XML data by csv data. (imported with a post_init hook). Thus as we now have a csv file it can be imported throught the web interface using the CSV import tool. This should be enough to replace the use of the update button in settings. Can you confirm?

guewen:
:+1:

BT-ojossen:
:+1: seems to be fine to me

PR #186, opened 2015-10-29
Print bank address instead company partner address

paulius-sladkevicius:
We've faced that company address can be different from bank account, so here is suggestion
to use res.partner.bank address instead company partners
e.g. Company is based in Belgium, but bank account is from Switzerland

### DIFF
--- a/l10n_ch_payment_slip/bank.py
+++ b/l10n_ch_payment_slip/bank.py
@@ -17,7 +17,7 @@
 #    along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 ##############################################################################
-from openerp import models, fields
+from openerp import models, fields, api
 
 
 class ResPartnerBank(models.Model):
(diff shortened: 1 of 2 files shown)


yvaucher:
I still don't get the issue here. Could you provide an example of printout before and after ?

Plus what kind of account is it for? postal account or bank account?

paulius-sladkevicius:
@yvaucher after one year hardly will find where I've patched it.
However, we needed to show bank address instead company because company was located in other country, but payslip need to be for Switzerland bank. 

If you think that this is incorrect way (worked for us), you can cancel it.

yvaucher:
Is what you need a BESR / BVRB or an ESR / BVR ?

BESR / BVRB:
https://www.ubs.com/ch/en/swissbank/corporates/cash-management/domestic-abroad/switzerland/_jcr_content/par/linklist/link_0.1979077066.file/bGluay9wYXRoPS9jb250ZW50L2RhbS91YnMvY2gvc3dpc3NiYW5rL2NvcnBvcmF0ZXMvY2FzaC1tYW5hZ2VtZW50L3BheW1lbnQtc2VydmljZXMvZG9jdW1lbnRzL2hhbmRidWNoLWJlc3ItZW4ucGRm/handbuch-besr-en.pdf

I don't think your PR is valid for both.

yvaucher:
Thus I'll close this, thanks for sharing it anyway this is much appreciated and sorry that nobody was interested to review it earlier. Feel free to reopen it or make plenty of other PR and ping me for a slighly fastest reaction ;-) .

PR #194, opened 2015-11-17
Fix module not installabe and relation table name

ecino:
Small fixes to let install module lsv_dd.


yvaucher:
Thanks :+1:

yvaucher:
I fast track this as the module is unusable without this.

PR #195, opened 2015-11-18
Missing security group in partner view of l10n_ch_scan_bvr added

tj-jamotion:
Missing security group in partner view of l10n_ch_scan_bvr added
Otherwise users without this right can't access contacts.

### DIFF
--- a/l10n_ch_scan_bvr/views/partner_view.xml
+++ b/l10n_ch_scan_bvr/views/partner_view.xml
@@ -6,7 +6,7 @@
       <field name="model">res.partner</field>
       <field name="inherit_id" ref="base.view_partner_form" />
       <field name="type">form</field>
-      <field name=groups_id" eval="[(4, ref('account.group_account_invoice'))]" />
+      <field name="groups_id" eval="[(4, ref('account.group_account_invoice'))]" />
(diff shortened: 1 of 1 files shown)


yvaucher:
In fact you need at least base.group_user ("Employee")

Nevertheless, it seams completely reasonable to show only this for the Invoice & payment group.

Thanks for the fix. :+1: 

A little tip, in such case you can do `git commit --amend` to change your last commit for a fixup.

yvaucher:
@tj-jamotion If you have already signed the CLA feel free to ping so we can check it.

l10n_ch_scan_bvr/views/partner_view.xml
```diff
@@ -6,6 +6,7 @@
       <field name="model">res.partner</field>
       <field name="inherit_id" ref="base.view_partner_form" />
       <field name="type">form</field>
```
yvaucher, line 8: By the way, you can remove this line, as it is deprecated in 8.0

jgrandguillaume:
Hi,

I confirm @tj-jamotion  signed the CLA.
:+1: 

Thanks for the contribution,

Regards,

PR #197, opened 2015-11-20
8.0

johannbelet:
This fix removes the leading space in BVR string. This space can cause
compatibility issues with some BVR scanners.


tafaRU:
@johannbelet, the target branch is wrong, it should be 8.0 instead 9.0.

PR #198, opened 2015-11-26
[FIX] call to moved methods

ecino:
Code was still referencing olg methods that were moved into another file. This fix is mandatory for the module to work.

### DIFF
--- a/l10n_ch_lsv_dd/wizard/dd_export_wizard.py
+++ b/l10n_ch_lsv_dd/wizard/dd_export_wizard.py
@@ -199,30 +199,31 @@ class PostDdExportWizard(models.TransientModel):
         control_range = self._gen_control_range('47', properties)
 
         vals = collections.OrderedDict()
-        self._check_currency(line, properties)
+        export_utils.check_currency(line, properties)
         self._check_amount(line, properties)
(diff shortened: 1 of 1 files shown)


yvaucher:
:+1:

l10n_ch_lsv_dd/wizard/dd_export_wizard.py
```diff
@@ -199,30 +199,30 @@ def _generate_debit_record(self, line, properties, payment_order):
...
-            ('reserve_11', self._complete_line('', 3)),
-            ('reserve_12', self._complete_line('', 1)),
-            ('orderer_info', self._complete_line('', 140)),  # Not implemented
-            ('reserve_13', self._complete_line('', 14))
+            ('reserve_6', export_utils.complete_line(35)),
+            ('reserve_7', export_utils.complete_line(35)),
+            ('reserve_8', export_utils.complete_line(35)),
+            ('reserve_9', export_utils.complete_line(10)),
+            ('reserve_10', export_utils.complete_line(25)),
+            ('communication', export_utils.complete_line(140, communications)),
```
yvaucher: This line is too long. It doesn't pass pylint test

PR #219, opened 2016-02-29
Update de.po

BT-ojossen:


leemannd:
I think we can close this PR. It is the same modification as #218 
:-1:

jgrandguillaume:
As @leemannd said, I close this one.

PR #208, opened 2016-02-02
return list of all banks if name is empty

BT-ojossen:
otherwise the customer thinks that there aren't any banks in the system.

PR #226, opened 2016-04-05
return list of all banks if name is empty

BT-fgarbely:
(same like pull request #208)


yvaucher:
@BT-fgarbely You should do a cherry-pick of the commit from @BT-ojossen 

```
git cherry-pick b0f970baf010435f3e3eff72e6390a3d2105cf
```

The author is incorrect

PR #227, opened 2016-04-07
return list of all banks if name is empty

BT-fgarbely:
otherwise the customer thinks that there aren't any banks in the system.


BT-fgarbely:
This pull request replaces #226

yvaucher:
@BT-fgarbely Thanks :+1: 

(Note that you can do a `git push -f` to force update of a PR

PR #233, opened 2016-04-25
Add missing view to __openerp__.py

dcoopennet:


yvaucher:
:+1:

yvaucher:
As it is trivial I fast track it

PR #234, opened 2016-05-03
Fix test by separating l10n_ch_lsv_dd as it pulls account_banking_pay…

yvaucher:
…ment_export which adds a required type field on payment.mode

The required field is added here:

https://github.com/OCA/bank-payment/blob/bafffd0eded6a7eb71570d3f22bfb91dc19ba4bc/account_banking_payment_export/models/payment_mode.py#L70-L71

```
l10n_ch_lsv_dd
└── account_banking_mandate
    └── account_banking_payment_export
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,10 @@ env:
   matrix:
   - LINT_CHECK="1"
   - TRANSIFEX="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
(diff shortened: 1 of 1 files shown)


yvaucher:
@ecino
l10n_ch_dta demo data failing were due to l10n_ch_lsv_dd dependencies.
Hoping to get tests green.

PR #238, opened 2016-05-17
Fix bug in ubs parser

ecino:


yvaucher:
Tests are failing on this:

```
2016-05-17 07:55:37,180 24058 ERROR openerp_test openerp.addons.l10n_ch_account_statement_base_import.parsers.base_parser: Error when parsing ubs_csvparser file

Traceback (most recent call last):

File "/home/travis/build/OCA/l10n-switzerland/l10n_ch_account_statement_base_import/parsers/base_parser.py", line 61, in parse

return self._parse()

File "/home/travis/build/OCA/l10n-switzerland/l10n_ch_account_statement_base_import/parsers/ubs_file_parser.py", line 157, in _parse

'transactions': self._parse_transactions(),

File "/home/travis/build/OCA/l10n-switzerland/l10n_ch_account_statement_base_import/parsers/ubs_file_parser.py", line 123, in _parse_transactions

amount = - float_or_zero(line[debit]) or \

KeyError: u'D\xc3\xa9bit'
```

ecino:
I think I changed the encoding without noticing... I will check, thanks.

l10n_ch_account_statement_base_import/parsers/ubs_file_parser.py
```diff
@@ -118,8 +118,8 @@ def _parse_transactions(self):
                 line.get("Description 3", '')]
             label = ' '.join(filter(None, descriptions))
 
-            debit = "D�bit".decode('iso-8859-15')
-            credit = "Cr�dit".decode('iso-8859-15')
+            debit = "Débit".decode('iso-8859-15')
+            credit = "Crédit".decode('iso-8859-15')
```
yvaucher: I don't see what has changed here. Strange

yvaucher: Indeed it must be some encoding change. As we see no difference

file old_file.py 
old_file.py: Python script, ISO-8859 text executable, with CRLF line terminators

file new_file.py 
new_file.py: Python script, UTF-8 Unicode text executable, with CRLF line terminators

So the new file is correct.

You might just want to put:

 u"Débit".encode('iso-8859-15')

l10n_ch_account_statement_base_import/parsers/ubs_file_parser.py
```diff
@@ -118,8 +118,8 @@ def _parse_transactions(self):
                 line.get("Description 3", '')]
             label = ' '.join(filter(None, descriptions))
 
-            debit = "D�bit".decode('iso-8859-15')
-            credit = "Cr�dit".decode('iso-8859-15')
+            debit = u"Débit".decode('iso-8859-15')
+            credit = u"Crédit".decode('iso-8859-15')
```
yvaucher: Shouldn't it be `encode` instead of `decode` ? Source is unicode and you want `iso-8859-15`

This will raise 

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 1: ordinal not in range(128)
```

ecino:
Sorry, I fixed a bit too quickly. Let's see if tests will pass now.

yvaucher:
@ecino 

It's a bit better as it doesn't fail on decode, however it's still failing.

```
openerp.exceptions.Warning: b'The following problem occurred during ubs_csvparser import. The file might not be valid.\n\n D\xe9bit'
```

And we finally can understand where is the issue.

I think that the `decode` was only there because of the wrong format of python file. It seems that the `csv` file is already being parsed by decoding from `iso-8859-15` at [ubs_file_parser.py:63](https://github.com/OCA/l10n-switzerland/blob/8.0/l10n_ch_account_statement_base_import/parsers/ubs_file_parser.py#L63) which transforms the key and values to `unicode`

Thus, the right answer is to simply have `unicode` values:

``` python
           debit = u"Débit"
           credit = u"Crédit"
```

ecino:
Thank you for your help! fingers crossed :wink:

yvaucher:
It's red. Buuuut it only lint errors now. Nicely done. The test are good.

And it seems the lint tool is failing.

Already my :+1: here

yvaucher:
In case you missed it, in your last changes you introduced: 

```
l10n_ch_fds_postfinance/wizards/fds_files_import_tobankstatments_wizard.py:120:29: F841 local variable 'e' is assigned to but never used
```

l10n_ch_account_statement_base_import/data/UBS_export.csv
```diff
@@ -7,4 +7,5 @@ Date d'
 31.08.2014;0298 00101368;;0298 00101368.01Q;CH08 0029 8298 9999 9901 Q;CHF;01.08.2014;31.08.2014;UBS Current Account for Private Clients;10.08.2014;31.08.2014;10.08.2014;Dividend;VN   472672     101368S3;NOKIA;BC80538KE0773604;;;;76.00;20'365.60
 
 Solde final;Solde initial
-20827.000000000;20289.600000000
\ No newline at end of file
+20827.000000000;20289.600000000
```
yvaucher, line 10: Here I definitely think the error is not this file but the lint tool. I wonder if this caused the travis error:

2016-07-29 07:18:29,182 24073 ERROR openerp_test openerp.addons.l10n_ch_account_statement_base_import.tests.test_ubs: ` openerp.exceptions.Warning: The following problem occurred during ubs_csvparser import. The file might not be valid.

list index out of range might be due to the new line as parser tries to read an additional line?

ecino:
Ok, now it works. I should also import module wand in the runbot but I don't know how to do this?

ecino:
Any status update on this?

yvaucher:
Merged :smile:

PR #240, opened 2016-05-17
[8.0] l10n_ch_scan_bvr - fix payment term

gregorlazzarato:
When creating the invoice from esr scan the customer payment term is used instead of supplier payment term

### DIFF
--- a/l10n_ch_scan_bvr/wizard/scan_bvr.py
+++ b/l10n_ch_scan_bvr/wizard/scan_bvr.py
@@ -212,7 +212,8 @@ class ScanBvr(models.TransientModel):
                  'ccp': data['bvr_struct']['beneficiaire']})
         date_due = today
         # We will now compute the due date and fixe the payment term
-        payment_term_id = account_info.partner_id.property_payment_term.id
+        payment_term_id = (account_info.partner_id.
+                           property_supplier_payment_term.id)
         if payment_term_id:
             # We Calculate due_date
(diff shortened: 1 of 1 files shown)


yvaucher:
Can you amend your commit to give it a better name than bugfixing plz.

Plus fix the following travis lint error:

```
l10n_ch_scan_bvr/wizard/scan_bvr.py:205:80: E501 line too long (83 > 79 characters)
```

renzo1974:
Hi
I have updated this PR now with my user and added a line break for respecting the OCA rules...
But I do not understand why you stil use only 79 chars per line... 
Odoo has officially commited that the "[PEP8 E501: line too long"](https://www.odoo.com/documentation/8.0/reference/guidelines.html#python) can be ignored (in Documentation for Odoo 8.0). So we are using the rules provided by Odoo. All our checks are based on these rules.
Regards
Renzo

yvaucher:
@rm-jamotion Thanks for the change

About limitation to 79 characters the idea is to be able to have on the same screen 2 page of code side to side. We had, some years ago, a vote on 79 or 99 characters and the vote resulted to 79 characters so people with smaller screens like on laptop would still be able to code.

Odoo has never enforced a line size limitation. And while at last Odoo start to respect PEP8, OCA enforced it quicker. OCA coding guideline are different from Odoo's ones.

https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#differences-with-odoo-guidelines

And you have my :+1:

Issue #243, opened 2016-05-27
Account Type not correctly detected


user1007017:
Hi, I'm getting the following error:

> The bank account 01-234567-8 used in invoice needs to be a postal account or have a bank CCP.

Installed Odoo 9 and the following "apps"
- Switzerland - Bank list   Camptocamp, brain-tec AG, Odoo Community Association (OCA)  Localisation    9.0.1.0.0
- Switzerland - Bank type   Camptocamp, Odoo Community Association (OCA)    Localization    9.0.1.0.0
- Switzerland - Payment Slip (BVR/ESR)  Camptocamp, Odoo Community Association (OCA)    Localization    9.0.2.1.0

The account number entered at Contacts / Bank Accounts / Bank Accounts was discovered correctly as Postfinance
But after clicking the BVR/ESR button, the message appears

dcoopennet:
I think the issue came from this double negativ check here.
" not " +  " != "
if not bank_acc.acc_type != 'postal' or not bank_acc.ccp:
                    msg.append(_('The bank account {} used in invoice needs to'
                                 ' be a postal account or have a bank CCP.'
                                 ).format(bank_acc.acc_number))
Please fix this as soon as possible.

dcoopennet:
This issue makes the module not working. Please fix this. @yvaucher

yvaucher:
@dcoopennet Can you do a pull request ?

yvaucher:
Fixed by #245

PR #245, opened 2016-06-03
Fix issue #243 with double negative

dcoopennet:
@yvaucher As asked this is the PR about the issue #243.

### DIFF
--- a/l10n_ch_payment_slip/models/invoice.py
+++ b/l10n_ch_payment_slip/models/invoice.py
@@ -158,7 +158,7 @@ class AccountInvoice(models.Model):
                     msg.append(_('The bank account {} used in invoice has no '
                                  'BVR/ESR adherent number.'
                                  ).format(bank_acc.acc_number))
-                if not bank_acc.acc_type == 'postal' or not bank_acc.ccp:
+                if not (bank_acc.acc_type == 'postal' or bank_acc.ccp):
                     msg.append(_('The bank account {} used in invoice needs to'
                                  ' be a postal account or have a bank CCP.'
                                  ).format(bank_acc.acc_number))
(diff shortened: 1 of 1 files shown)


l10n_ch_payment_slip/models/invoice.py
```diff
@@ -158,7 +158,7 @@ def _check_bvr_generatable(self):
                     msg.append(_('The bank account {} used in invoice has no '
                                  'BVR/ESR adherent number.'
                                  ).format(bank_acc.acc_number))
-                if not bank_acc.acc_type != 'postal' or not bank_acc.ccp:
+                if not bank_acc.acc_type == 'postal' or not bank_acc.ccp:
```
yvaucher: Okay this is not exactly what we want.

Before the fix: only bank account with ccp works.
After the current fix only account of type postal will work.

Here are the 3 use cases:
- A: bank account without ccp -> raise an error
- B: bank account with ccp -> OK
- C: account type postal (thus always with ccp) -> OK

Only A must fail.

Thus the condition must be 

`bank_acc != 'postal and not bank_acc.ccp'`

unless you prefer:

`not (bank_acc.acc_type == 'postal' or bank_acc.ccp)`

Or even we could simplify it to `not bank_acc.ccp` as ccp should be computed based on account_number if the type is postal anyway.
https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_base_bank/models/bank.py#L187

yvaucher:
Thanks :+1:  Tested on runbot + code review

yvaucher:
As this is blocking I'll fastforward it. Merging...

yvaucher:
I merged it (and also added you as contributor)

dcoopennet:
@yvaucher Thanks a lot for your fast reaction. It's a pleasure to interact with you.

PR #156, opened 2015-07-27
Winbiz port

vrenaville:
This module implement to import file from Winbiz Salaire.

PR #253, opened 2016-07-15
9.0 cresus import

open-net-sarl:
I basically reenabled the module on 9.0 and fixed what came up. And added tests. I also had to show `BaseModel.load` the door because it matches account codes too broadly (proof: [Travis results when you revert that change](https://travis-ci.org/open-net-sarl/l10n-switzerland/builds/144994088)). While at it, I improved the UI, just like [this comment](https://github.com/OCA/l10n-switzerland/pull/156#issuecomment-135695263) suggested on a sister module (which I'll be done porting too soon). More significantly: I  made it so that the module could import more complicated moves, which Crésus indicates with an ellipsis in the account code columns, which also prompted me to group lines into moves based on the “pce” number rather than the date.

For the record: I had data coming from Crésus Comptabilité 3.3m PRO to test around, and that's why I came across ellipses. I know the module was designed for Crésus Salaires. Anyway, it's probably better that way. I didn't encounter lines with negative “amount”, but it was already supported in the 8.0 module, so I kept it.


l10n_ch_import_cresus/__openerp__.py
```diff
@@ -1,36 +1,24 @@
...
+# Copyright 2015 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {
     'name': 'Account Import Cresus',
-    'version': '8.0.1.0.0',
+    'summary': 'Allows to import Cresus Salaires .txt files containing \
+                journal entries into Odoo.',
+    'version': '9.0.1.0.0',
+    'category': 'uncategorized',
```
yvaucher: What about "Human Resources" as category ?

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
-        return (values)
+        for format in ['%d.%m.%y', '%d.%m.%Y']:
+            try:
+                dt = datetime.strptime(date_string, format)
+                break
+            except ValueError:
+                continue
+        else:
+            raise exceptions.ValidationError(
+                "Can't parse date '%s'" % date_string)
```
yvaucher: The error message should be translatable.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
+        return fields.Date.to_string(dt)
 
-    def _manage_load_results(self, result):
-        """Manage the BaseModel.load function output and store exception.
+    def _find_account(self, code):
+        account_obj = self.env['account.account']
+        account = account_obj.search([('code', '=', code)], limit=1)
+        if not account:
+            raise exceptions.MissingError("No account with code %s" % code)
```
yvaucher: Error message to wrap with `_( ... )`

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
+        return account
 
-        Will generate success/failure report and store it into report field.
-        Manage commit and rollback even if load method uses PostgreSQL
-        Savepoints.
+    def _find_tax(self, typtvat, account):
+        tax_obj = self.env['account.tax']
+        if not typtvat or account.user_type_id.include_initial_balance:
+            return tax_obj.browse()
```
yvaucher: no need to call the `browse()`

`tax_obj` is already a record set.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
-
+                try:
+                    data = csv.DictReader(decoded, fieldnames=self.HEAD_CRESUS,
+                                          delimiter=delimiter)
+                except csv.Error as error:
+                    raise exceptions.ValidationError('''\
+CSV file is malformed
+Please choose the correct separator
+the error detail is:
+%r''' % error)
```
yvaucher: Why breaking the indent ? The format:

```
    (" ... "
     " ...")
```

is  valid.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -1,131 +1,65 @@
...
+    @staticmethod
+    def make_move(lines, **kwargs):
+        # assert set(kwargs.keys()) == self.ODOO_MOVE_ARGS
+        kwargs.update({'line_ids': [(0, 0, ln) for ln in lines]})
+        return kwargs
 
-        :param messages: return of BaseModel.load messages key
-        :returns: formatted string
+    ODOO_LINE_ARGS = {'account_id', 'partner_id', 'name',
+                      'tax_line_id', 'analytic_account_id'}
```
yvaucher: This isn't used, to remove ?

l10n_ch_import_cresus/i18n/fr.po
```diff
@@ -1,163 +1,202 @@
...
-msgid "Import Ecritures"
-msgstr "Import écritures"
+msgid "Import Move lines"
+msgstr "Import Move lines"
 
 #. module: l10n_ch_import_cresus
-#: field:account.cresus.import,help_html:0
+#: model:ir.ui.view,arch_db:l10n_ch_import_cresus.cresus_move_line_importer_form
+msgid "Import file"
+msgstr "Import file"
```
yvaucher: Keep the string empty when it's not translated.

l10n_ch_import_cresus/tests/test_import.py
```diff
@@ -0,0 +1,78 @@
...
+        super(TestImport, self).setUp()
+
+        tax_obj = self.env['account.tax']
+        account_obj = self.env['account.account']
+        analytic_account_obj = self.env['account.analytic.account']
+
+        user_type = self.env['account.account.type'].search(
+            [('include_initial_balance', '=', False)], limit=1)
+        for code in '1000 1010 1210 2200 2800 2915 6512 6513 6642 '\
+                    '9100 10101'.split():
```
yvaucher: Why using a string instead of a list ?

l10n_ch_import_cresus/tests/test_import.py
```diff
@@ -0,0 +1,78 @@
...
+07.01.2002	6642	1010	1	Frais de déplacement - Frauenfeld	45.00
+07.01.02	6513	1010	2	Frais de prospection - Tartempion	218.50
+12.01.02	1010	6642	3	Frais de déplacement - Tolochenaz	-15.00
+14.01.02	6512	...	4	Salt, (TVA) net, TVA = 13.17	173.38
+14.01.02	2200	...	4	Salt, 7.6% de TVA (TVA)	13.17	VAT
+14.01.02	...	1010	4	Salt Total, (TVA)	186.55
+''')})
+        wizard.import_file()
+
+        res = move_obj.browse(wizard.imported_move_ids.ids)
```
yvaucher: `res = move_obj.browse(wizard.imported_move_ids.ids)` => `res =  wizard.imported_move_ids`

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -1,131 +1,65 @@
...
-            rows = msg.get('rows', {})
-            res.append(_("%s. -- Field: %s -- rows %s to %s") % (
-                msg.get('message', 'N/A'),
-                msg.get('field', 'N/A'),
-                rows.get('from', 'N/A'),
-                rows.get('to', 'N/A'))
-            )
-        return "\n \n".join(res)
+    @staticmethod
+    def make_line(debit_amount, credit_amount, **kwargs):
```
yvaucher: `prepare_line` would be more suited

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
+                    lines,
+                    date=previous_date,
+                    ref=previous_pce,
+                    journal_id=journal_id)
+                lines = []
+            previous_pce = line_cresus['pce']
+            previous_date = line_cresus['date']
+
+            recto_amount = float(line_cresus['amount'].replace('\'', '')
+                                                      .replace(' ', ''))
```
yvaucher: Maybe you could use

```
import locale

locale.setlocale(locale.LC_ALL, 'fr_CH')
locale.atof(line_cresus['amount'])
```

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
+                tax = self._find_tax(line_cresus['typtvat'], account)
+                analytic_account = self._find_analytic_account(
+                    line_cresus['analytic_account'], account)
+                lines.append(self.make_line(
+                    recto_amount, verso_amount,
+                    account_id=account.id,
+                    partner_id=False,
+                    name=line_cresus['ref'],
+                    tax_line_id=tax.id,
+                    analytic_account_id=analytic_account.id))
```
yvaucher: don't you want to pass the account to the method "make_line" so you can define all those arguments in it ?

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -137,194 +71,150 @@ def _parse_csv(self):
...
+        self.imported_move_ids = [move_obj.create(mv).id for mv in data]
 
     @api.multi
     def import_file(self):
-        data = self._parse_csv()
-        new_data = self._standardise_data(data)
-        return self._load_data(new_data)
+        try:
+            self._import_file()
+        except Exception as exc:
```
yvaucher, line 192: Could you provide a more precise type of exception to catch ?

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus_view.xml
```diff
@@ -1,48 +1,43 @@
...
-            	<field name="help_html"/>
-            </page>
-          </notebook>
-        </form>
-      </field>
+        <field name="name">cresus move line importer</field>
+        <field name="model">account.cresus.import</field>
+        <field name="arch" type="xml">
+            <form string="Move / Move lines importer">
+                <footer>
```
yvaucher: Could you move this footer at the bottom would make more sense ?

yvaucher: Don't we want also a cancel button ?

yvaucher:
@lbeopennet Hello thanks for having added some tests.

Please see my comments

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
ghost: does it also need `_( .... )` ?

l10n_ch_import_cresus/tests/test_import.py
ghost: tbh the main reason is that I was too lazy to add a comma manually. It does the same in less space tho, and it echoes Ruby's `%w,1000 1010 1210 2200,` quite nicely IMHO

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
yvaucher: Indeed it does for all Odoo exceptions.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
ghost: fr_CH doesn't consider the space as a valid thousands separator so it won't do the same, and I essentially got it this way from 8.0. I'd rather keep it in this less abstracted way because I'm not sure exactly when either of those occur.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
ghost, line 192: The point is to catch any unexpected error and show it in our interface rather that the default "Odoo Server Error" popup. Got it this way from 8.0.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus_view.xml
ghost: Either nothing has been imported and I can just close the popup (It's all or nothing because of `cr.rollback`) or the import has succeeded and I'm automatically sent to a view on the moves I just imported, and I get a chance to manually delete some or all of them. The 'automatically sent' part was my idea tho, there was a IMHO-redundant 'view imported moves' button before so i can revert that if needs be.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
yvaucher: In fact it is a bad idea to use setlocale as it could change it for all threads

But you should use the right tool.

Thus it would be better to use pybabel (thanks @gurneyalex for the tip)

``` python
from babel.numbers import parse_decimale

float(parse_decimal("       1'099.98      ", locale='de_CH'))
```

l10n_ch_import_cresus/tests/test_import.py
yvaucher: Here you go to save your fingers:

```
['1000', '1010', '1210', '2200', '2800', '2915', '6512', '6513', '6642', '9100', '10101']
```

yvaucher:
About translations, you should probably exports all translations (after having added all `_( ... )`  to error messages) in a template (Developper mode / Settings / Translations / Export .
Use new language and select this module. This should create you a `.pot` file.

Thus use a po editor. For example poedit (free opensource) and open the `.fr` file and update it with the `.pot` file you just created. Thus you can edit in the GUI. Otherwise you can still use gettext to do the update in commandline.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
ghost: Okay that'll do

ghost:
About translations, been here done that, but when I do that, even on a test database, I get back a file that stats with this ghost translation:
`#. module: l10n_ch_import_cresus
#: code:addons/l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py:64
#, python-format
msgid "\n"
"                 In order to import your 'Cresus Salaires' .txt                  file you must complete the following requirements : </br>\n"
"                * The accounts, analytical accounts used in the Cresus                 file must be previously created into Odoo  </br>\n"
"                * The date of the entry will determine the period used                 in Odoo, so please ensure the period is created already. </br>\n"
"                * If the Cresus file uses VAT codes (i.e: IPI),                 please make sure you have indicated this code in the                 related Odoo tax (new field).                 Warning, the Odoo tax must be 'tax included'.                 If the tax does not exist you have to create it. </br>\n"
"                * All PL accounts must have a deferral method = 'none'                (meaning: no balance brought forward in the new fiscal year)                and all\n"
"                 Balance sheet account must have a deferral                 method = 'balance'. </br>"
msgstr ""`
which is just not [there](https://github.com/open-net-sarl/l10n-switzerland/blob/9.0-cresus-import/l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py#L64) ! what's wrong? (I know you won't necessarily guess, maybe it's on our server, and I have to bounce anyway, I'll try on the runbot tomorrow and then come back to complain)

ghost:
(still testing, hold on)

ghost:
ok, so I got around all of that, the runbot was immeasurably helpful. Hey @yvaucher, mind reviewing me again?

l10n_ch_import_cresus/models/account_tax.py
```diff
@@ -0,0 +1,13 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from openerp import models, fields
+
+
+class AccountTax(models.Model):
+    _inherit = "account.tax"
+
+    tax_cresus_mapping = fields.Char(
+        size=128,
```
yvaucher: Just asking, do you know if the size 128 is a limitation of cresus ? if not size attribute can be dropped.

l10n_ch_import_cresus/tests/cresus.txt
```diff
@@ -0,0 +1,12 @@
...
+01.01.02	1010	9100		Solde à nouveau CCP	8'000.00			10
+01.01.02	1210	9100		Solde à nouveau Matériel	100'000.00			10
+01.01.02	9100	2800		Solde à nouveau Capital	20'000.00			10
+01.01.02	9100	2915		Solde à nouveau Réserve générale	323.20			10
+07.01.2002	6642	1010	1	Frais de déplacement - Frauenfeld	45.00
+07.01.02	6513	1010	2	Frais de prospection - Tartempion	218.50
+12.01.02	1010	6642	3	Frais de déplacement - Tolochenaz	-15.00
+14.01.02	6512	...	4	Salt, (TVA) net, TVA = 13.17	173.38
+14.01.02	2200	...	4	Salt, 7.6% de TVA (TVA)	13.17	VAT
+14.01.02	...	1010	4	Salt Total, (TVA)	186.55
```
yvaucher, line 11: Thanks that's cleaner to have an exemple file on it's own :+1:

l10n_ch_import_cresus/models/account_tax.py
ghost: I have no idea. Besides, maybe it's only the case on earlier versions, so better drop it.

l10n_ch_import_cresus/tests/cresus.txt
ghost, line 11: Thanks ^^

yvaucher:
@lbeopennet Thanks for all the changes LGTM now

:+1:

l10n_ch_import_cresus/models/account_tax.py
yvaucher: @lbeopennet you are right, I just checked with @vrenaville and it can be dropped. Size was mandatory in older version of Odoo and there is no reason to keep it.

leemannd:
:+1:

PR #255, opened 2016-08-08
Add a module to import WinBIZ accounting

ghost:
This module contains a wizard that allows the user to import Excel files containing journal items that have been generated by WinBIZ. It is based on a similar module by @vrenaville for Odoo 8.0, which was never merged (PR #156).

I changed the input format from XML to Excel in order to try to support WinBIZ Basic, which doesn't include XML exporting as a feature. This creates an external dependency on xlrd. I added it with pip to Travis, but maybe that's not how it's done.

In order to write a unit test, I also made up a quick and dirty format for dumping the contents added to the table account.move and account.move.line by the importing process, which is basically the output of the import function. The test works by running the import process with some sample input, and then comparing its output with the expected “golden” output that accompanies the current commit. It may seem a bit self-serving as a test given that the golden output is ultimately a past output from the program itself, but it allowed me to see instantly exactly what my changes to the code would result in in terms of output and it will notice if a change to another module messes its normal operation, so I think it's a good unit test.

I also improved the UI a bit, as the functional tests suggested in #156, and replaced `Model.load` for the same reason as in #253.


l10n_ch_import_winbiz/README.rst
```diff
@@ -0,0 +1,71 @@
...
+* The export encoding should be “Windows Ansi - 1252” (default)
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/l10n-switzerland/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smashing it by providing a detailed and welcomed feedback
+`here <https://github.com/OCA/l10n-switzerland/issues/new?body=module:%20l10n_ch_import_winbiz%0Aversion:%208.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
yvaucher: You can remove this link to matches with new readme template:

https://raw.githubusercontent.com/OCA/maintainer-tools/master/template/module/README.rst

yvaucher:
`I changed the input format from XML to Excel in order to try to support WinBIZ Basic`

@lbeopennet Thus shouldn't we keep both format ? Excel format has some limits.

l10n_ch_import_winbiz/wizard/l10n_ch_import_winbiz.py
```diff
@@ -0,0 +1,289 @@
...
+        else:
+            return datetime.datetime(*xldate_as_tuple(date, self.wb.datemode))
+
+    @api.multi
+    def _standardise_data(self, data):
+        """
+        This function split one line of the spreadsheet into multiple lines.
+        Winbiz just writes one line per move.
+        """
+
```
yvaucher, line 96: That's a huge function. Can't it be splitted ? Can't the inner class be declared outside ? what is the point of declaring it there ?

ghost:
@yvaucher XML is now supported again, as I splitted the format-specific bits into classes and put them in `utils.importers`. The mashing-lines-together part is still in `wizard` but I dedented most of it, since it doesn't need to be nested inside the scope of `_standardise_data` (although I wouldn't move them out when I wrote them as I wasn't sure they would never need it). There's still `find_account`, which I could write as a method on `AccountWinbizImport`, but I find it prettier that way :grin:

cyp-opennet:
LGTM :+1:

yvaucher:
@lbeopennet Thanks for the changes :+1:

PR #256, opened 2016-08-09
9.0 cresus import bis

ghost:
These are just some simple fixes to l10n_ch_import_cresus, which I've previously ported to 9.0 (#253) before touching another similar module, l10n_ch_import_winbiz, which is still in PR (#255). Working again on the same kind of things made me learn some things I didn't know the first time, so I had to backport a bit. I also took the opportunity to fix some minor details in the README.rst.


l10n_ch_import_cresus/tests/test_import.py
```diff
@@ -42,32 +47,63 @@ def setUp(self):
...
+        if not misc:
+            misc = journal_obj.create({
+                'code': 'MISC',
+                'name': 'dummy MISC',
+                'type': 'general'})
+
+        def get_path(filename):
+            res = get_resource_path('l10n_ch_import_cresus', 'tests', filename)
+            return res
+        input = open(get_path('input.csv'))
```
yvaucher: Please don't use built-in `input` function name for your variable.

Let's say you remove this line later, you won't get an error of undefined variable.

yvaucher:
Just a tiny remark otherwise looks good to me. :+1:

l10n_ch_import_cresus/tests/test_import.py
ghost: I see. Also its scope is way bigger than it needs be. Gotta fix that.

l10n_ch_import_cresus/tests/test_import.py
```diff
@@ -42,32 +47,64 @@ def setUp(self):
...
-        self.assertEqual(res[2].date, '2002-01-07')
-        self.assertEqual(res[3].line_ids[0].account_id.code, '6642')
-        self.assertEqual(res[3].line_ids[0].debit, 15.0)
-        self.assertEqual(res[3].line_ids[0].credit, 0.0)
-        self.assertEqual(res[3].line_ids[1].account_id.code, '1010')
-        self.assertEqual(res[3].line_ids[1].debit, 0.0)
-        self.assertEqual(res[3].line_ids[1].credit, 15.0)
-        self.assertEqual(res[4].line_ids[1].tax_line_id, self.vat)
+
+        gold = open(get_path('golden-output.txt'))
```
yvaucher, line 75: This file is opened but not closed, you might want to use `with` syntax or call `close()`

l10n_ch_import_cresus/models/account_tax.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2015 Camptocamp SA
+# Copyright 2016 Open Net Sàrl
```
yvaucher: You can add your copyright but you cannot remove existing one, it is AGPL infrigement event when you provide those modules to your customer, you have to respect the AGPL license.

l10n_ch_import_cresus/wizard/l10n_ch_import_cresus.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright 2015 Camptocamp SA
+# Copyright 2016 Open Net Sàrl
```
yvaucher: This is copyright stealing you can not do that.

First version of the module is there:
https://github.com/OCA/l10n-switzerland/commit/09c366359b157afb3b85a0684b7cc23d603624ef

yvaucher (changes requested):
The change seems okay, `include_initial_balance` is not used anymore.

However, I'm really disapointed that the copyright is changed. OpenNET has been there for a while and should know the rules and what AGPL is.

In fact your changes of copyright does not respect the section 4 of AGPL. You have to keep original copyright notices whenever the copyright would apply without those mentions in most of the countries. Whenever you redistribute this code you do have to keep the right notices. It's not the first time I have seen such changes in your pull requests. You have to stop doing that !

```
4. Conveying Verbatim Copies.

You may convey verbatim copies of the Program's source code as you receive it, in any medium, provided that you conspicuously and appropriately publish on each copy an appropriate copyright notice; keep intact all notices stating that this License and any non-permissive terms added in accord with section 7 apply to the code; keep intact all notices of the absence of any warranty; and give all recipients a copy of this License along with the Program.
```

ghost:
I'm sorry

yvaucher:
Almost.

This is done by 1 copyright per line in all OCA modules

Thus you should have:

```
# -*- coding: utf-8 -*-
# Copyright 2015 Camptocamp SA
# Copyright 2016 Open Net Sàrl
```

ghost:
fixed, thank you very much

ghost:
Is that also the case if the year is the same ?

```
# © 2016 Julien Coux (Camptocamp), Open Net Sàrl
```

or

```
# © 2016 Julien Coux (Camptocamp)
# © 2016 Open Net Sàrl
```

?

yvaucher:
Prefer `Copyright` in full letters instead of © to be more explicit. (at some point the community choose it was better to be explicit)

Thus it is the same, better to have one line per author.

yvaucher:
To be clear:

    # Copyright 2016 Julien Coux (Camptocamp)
    # Copyright 2016 Open Net Sàrl

Is the way to go

l10n_ch_import_cresus/tests/test_import.py
ghost: fixed by 6a0891e45e0c15b6829f5d8c8399d65e6247c2fd

ghost:
@yvaucher I request a new review.

Maybe a [rebase](https://github.com/open-net-sarl/l10n-switzerland/tree/9.0_cresus-import_rebase) is in order?  It should be strictly equivalent to [the current PR when merged](https://github.com/OCA/l10n-switzerland/tree/refs/pull/256/merge), while not polluting the commit graph. I also squashed fixup-type commits and deleted a commit-revert pair.

yvaucher (approved):

yvaucher:
Okay we are good, yes please squash those last commit before I merge.

ghost:
@yvaucher rebase done.

PR #258, opened 2016-08-29
Add index on invoice_id for account.move.line

mdietrichc2c:
This PR is to correct a performance issue I noticed with a customer with millions of move lines.

Basically, since the `payment.slip` object has a related field `invoice_id` defined as `move_line_id.invoice_id` , validating an invoice will launch a search for payment slips with that invoice ID, which then launches another search for move lines with this invoice ID.

However, the default `invoice_id` in the move line does not have an index ; this means that a large part of the invoice's validation is spent searching for those move lines. With an index on this field, it's almost instantaneous.


guewen:
:+1:

yvaucher:
:+1:

PR #260, opened 2016-09-20
[FIX] set the right amount on BVR Split in case of partial payment on…

vrenaville:
Set the right due amount on BVR Slip in case of a partial payment  credit control line.


yvaucher (approved):
:+1:

PR #262, opened 2016-09-22
DTA: Remove unsupported bank type 'bv' (RecordGt827)

ghost:
In V9 res.partner.bank type 'bv' and 'bvr' was removed. instead a new acc type: 'postal' was introduced.
When creating a DTA now, every 'postal' bank is treated like a 'bvr', because it is the more common one. 

Therefore the old acc_type 'bv' (RecordGt827) is not supported any more


l10n_ch_dta/wizard/create_dta.py
```diff
@@ -724,10 +724,11 @@ def _create_dta(self, data):
                 record_type = RecordGt836
             elif country_code and country_code != 'CH':
                 record_type = RecordGt836
-            elif elec_pay == 'bvr':
+            #in v9 bv and bvr do not exist any more
+            elif elec_pay in ['bvr', 'postal']:
                 record_type = RecordGt826
-            elif elec_pay in ['bv', 'postal']:
-                record_type = RecordGt827
+            #elif elec_pay in ['bv', 'postal']:
```
yvaucher: Let's simply remove this, we have the git history for that.

l10n_ch_dta/wizard/create_dta.py
```diff
@@ -724,10 +724,11 @@ def _create_dta(self, data):
                 record_type = RecordGt836
             elif country_code and country_code != 'CH':
                 record_type = RecordGt836
-            elif elec_pay == 'bvr':
+            #in v9 bv and bvr do not exist any more
+            elif elec_pay in ['bvr', 'postal']:
```
yvaucher: AFAIK `elec_pay` can't be `'bvr'`. Simply check that it is equal to `'postal'` and remove `'bvr'`

yvaucher (changes requested):
Thanks for the fix. Can you make those little changes ?

l10n_ch_dta/wizard/create_dta.py
```diff
@@ -684,7 +684,7 @@ def _process_payment_lines(self, data, pline, elec_context, seq):
         number = number.replace('.', '').replace('-', '') or False
         elec_context['partner_bank_number'] = number
         elec_context['partner_bvr'] = ''
-        if pline.partner_bank_id.acc_type in ('bv', 'bvr', 'postal'):
+        if pline.partner_bank_id.acc_type == 'postal':
```
yvaucher: Second thinking about it:

This is not what we want.

Here it won't work with a postal number as IBAN or the CCP of the bank of your bank account.

As we can have a
- Postal account with a ccp number
- Postal account with an iban number
- Bank account for which the bank has a ccp number

In all those 3 case `ccp` will be filled with the ccp

```
- a postal account, we use acc_number
- a postal account in iban format, we transform acc_number
- a bank account with CCP on the bank, we use ccp of the bank
- otherwise there is no CCP to use
```

What we probably want to test here is if `pline.partner_bank_id.ccp` is set.

l10n_ch_dta/wizard/create_dta.py
```diff
@@ -724,10 +724,8 @@ def _create_dta(self, data):
                 record_type = RecordGt836
             elif country_code and country_code != 'CH':
                 record_type = RecordGt836
-            elif elec_pay == 'bvr':
+            elif elec_pay == 'postal':
```
yvaucher: Here we should cover all type of bank with ccp. This also require to change the first test on IBAN as an IBAN can be a postal account (IBAN will be more widely used in the future).

Maybe we need an option on payment object to choose if we want to generate a BV or BVR which is the payment mode choice instead of a bank account type.

yvaucher (changes requested):
Sorry I had to redive in the code change of l10n_ch_bank. Either we deprecate RecordGt827 or we could add an option to still allow to do it, maybe on the payment object.

We should also be aware of the possibility to accept bvrb (bvr with bank reference) which is widely used.

l10n_ch_dta/wizard/create_dta.py
yvaucher: More precisely you can check the field `communication_type` on `account.payment.line`. This field will be `bvr` for an elec_pay `bvr` and if it's `Free` it will be `bv`

ghost:
Hi @yvaucher 
Sorry the late reply. I added your recommended changes.

yvaucher:
The pylint failing is unrelated

yvaucher (approved):

yvaucher:
Thanks for the changes.

PR #263, opened 2016-09-23
[8.0] l10n_ch_payment_slip - show/hide parts of the partner and bank addresses

johannbelet:
Added the possibility to:
-Show/Hide Bank Zip and City
-Show full partner address or select only line 1 or 2 or hide address
-Show/Hide partner country
-Move amount
in payment_slip configuration (Company and Bank configs)

Added fr_CH translation


l10n_ch_payment_slip/company_view.xml
```diff
@@ -18,7 +18,11 @@
                             <field name="bvr_add_horz"/>
                             <field name="bvr_add_vert"/>
                             <field name="bvr_background"/>
+                            <field name="bvr_amount_line_horz"/>
+                            <field name="bvr_amount_line_vert"/>
                             <field name="merge_mode"/>
+	                    	<field name="bvr_print_cpny_address"/>
+							<field name="bvr_hide_cpny_country"/>
```
yvaucher: Please don't mix tab and spaces.

l10n_ch_payment_slip/i18n/fr_CH.po
```diff
@@ -0,0 +1,432 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+# 	* l10n_ch_payment_slip
```
yvaucher: Do we need a fr_CH.po ? We could simply use fr.po as when installing fr_CH it already use fr.po translations and override only the one specified in fr_CH.po

l10n_ch_payment_slip/payment_slip.py
```diff
@@ -480,10 +481,30 @@ def _draw_address(self, canvas, print_settings, initial_position, font,
...
+                    text.textLine(line)
+                    
+                text.textLine(city_line)
+                
+            if pt_cpny_add == 'second_line_only':
+                if com_partner.street2:
+                    line = com_partner.street2
+                    text.textLine(line)
+         
+                text.textLine(city_line)
```
yvaucher: This seems very specific, why would you like to display `street2` instead of `street` ?

l10n_ch_payment_slip/company.py
```diff
@@ -75,3 +87,17 @@ class ResCompany(models.Model):
         default="in_memory"
     )
     bvr_background = fields.Boolean('Insert BVR background ?')
+    
+    bvr_hide_cpny_country = fields.Boolean(
+        'BVR hide company country'
+    )
```
yvaucher: This field is never used please remove it.

yvaucher (changes requested):
Hello,

Thanks for the PR, could you explain a bit the need?

I can understand that the country is not required on a BVR. I think we can remove the country in any case, to do this we could skip the last line of `com_partner.contact_address.split("\n")` if country is defined.

``` python
address_lines = com_partner.contact_address.split("\n")
if com_partner.country_id:
    del address_lines[-1]
[...]

Instead of redoing what `res.partner` `_display_address` method is doing.

However, I struggling to understand the need of the use of only one of the `street` field. Maybe what you need is to be able to have a second company address only for BVR or invoicing.

For the bank zip, I think we need to add it anyway without a condition

Maybe an image of what is your expected result would help.
```

johannbelet:
Hi,

Thanks for the reply.

> I can understand that the country is not required on a BVR. I think we can remove the country in any case, to do this we could skip the last line of com_partner.contact_address.split("\n") if country is defined.
> 
> ```
> address_lines = com_partner.contact_address.split("\n")
> if com_partner.country_id:
>     del address_lines[-1]
> [...]
> ```
> 
> Instead of redoing what `res.partner` `_display_address` method is doing.

Yes indeed this should be done like that

l10n_ch_payment_slip/payment_slip.py
johannbelet: The purpose of this option is to be able to hide CP XXXX address line if there is one and just keep the address line (we were requested by our bank to remove CP number on BVR). But we want to keep CP number in our invoices.

l10n_ch_payment_slip/i18n/fr_CH.po
johannbelet: I added this one so you could use already translated new fields for fr.po
I was not aware of the override.

l10n_ch_payment_slip/payment_slip.py
yvaucher: So in your database all your addresses have CP in `street` and actual address in `street2` ?

This configuration is system wide whereas you could have 10k of customer address without CP and 7k of customer address with CP. How do you work with both?

As for you to work with this case you have to put the actual address in `street2` I guess that you identified addresses with CP. This makes `street` the "case postale" field. Wouldn't it be better to make a new type of address name `bvr` in which you don't put the CP? Otherwise make a boolean on the address to know if the first line must be skipped.
Yet I don't see the option to choose street or street2 at system wide a good generic option. Maybe you could do that change in a specific module for your client just redefining `_draw_address`

johannbelet: No we have usually Address in street and CP in street2 if there is any.
The original modifications offered to print "full address", "only line 1" (street) or "no address" but I added "only line 2" (street 2) just in case.

I guess it would be better to add a new bvr address but that would make information redondant in most cases. That is why I sorted it out like that.

I understand your point and yes I will add the street selection in a custom module.

johannbelet:
Ok,

I removed all the address selection part and the fr_CH translation.

l10n_ch_payment_slip/payment_slip.py
```diff
@@ -556,6 +564,8 @@ def _draw_bank(self, canvas, print_settings, initial_position, font, bank):
             if not line:
                 continue
             text.textLine(line)
+        line = (bank.zip + '' or '') + ' ' + (bank.city or '')
```
yvaucher: I guess you do `+ ''` to cast to string.

Prefer to use 

```
"%s %s" % (bank.zip or '', bank.city or '')
```

It's usually preferable instead of doing multiple concatenations: http://programmers.stackexchange.com/questions/304445/why-is-s-better-than-for-concatenation

However, in your case you might want to get rid of the extra space if one or the other is not defined.

Thus you can use the `join` method on a list of string:

```
line = [str(bank.zip or ''), bank.city]
line = ' '.join([s for s in line if s])
```

l10n_ch_payment_slip/company_view.xml
```diff
@@ -1,28 +1,30 @@
...
-                            <field name="bvr_add_vert"/>
-                            <field name="bvr_background"/>
-                            <field name="merge_mode"/>
-                        </group>
-                    </page>
-                </xpath>
-            </field>
-        </record>
-    </data>
+	<data>
```
yvaucher: By not mixing spaces and tab I meant we use spaces everywhere. Not tabs.

yvaucher:
@cyrilgdn this might interest you, plus #264 feel free to test, review

l10n_ch_payment_slip/payment_slip.py
johannbelet: My bad I didn't re-read my code... No I meant to write this:
`line = (bank.zip + ' ' or '') + (bank.city or '')`
But your solution is way better. Thank you for the very interesting link :)

yvaucher:
Thanks for the changes I squashed the commits, fixed the pep8 issues and merged it at 

https://github.com/OCA/l10n-switzerland/commit/8ef8f9b0646af068068951230754a0f96d1d4997

johannbelet:
Perfect thanks

yvaucher:
#275 for odoo 9.0

PR #264, opened 2016-09-23
[FIX] Allign Bank on address block

vrenaville:


yvaucher:
@vrenaville Maybe you can have a look at #263 ? It seems that there is duplicate work.

yvaucher:
Closing in favor of #263

leemannd:
@yvaucher Not the same target. Should we reopen this?

PR #267, opened 2016-09-29
[MINOR] add icons to our modules

ghost:
... so it looks better on Odoo Apps, mainly. Note that it reuses the icon that's already in `l10n_ch_hr_payroll` so the repository doesn't inflate, I checked.


angelmoya:
Checked on Runbot, works fine
:+1:

pedrobaeza:
I prefer to put a different icon per module, using the main one, but adding a subicon to specify the feature

PR #275, opened 2016-11-16
Hide partner country and added Horizontal and Vertical position to move

yvaucher:
amount

cherry-pick of #263


yvaucher:
included in https://github.com/OCA/l10n-switzerland/pull/303

PR #280, opened 2016-11-24
[9.0] l10n_ch_bank - Fix ccp computation from IBAN of postal account by normalizing it

yvaucher:
This allows to put IBAN with spaces separations for readability and visual check

ex CH09 0900 0000 1000 8060 7 becomes 10-8060-7


yvaucher:
travis error is fixed in #281

PR #282, opened 2016-11-25
Wrap xlrd import in try except to fix issue with loading module becau…

yvaucher:
…se of description image


yvaucher:
Fast tracking to fix the repo

PR #287, opened 2016-12-05
[PORT] Change installable to True

open-net-sarl:
No changes, just tested everything seems good.
Module base_location PR : https://github.com/OCA/partner-contact/pull/355

PR #290, opened 2017-01-02
[10.0][MIGRATION] Migrating l10n_ch_states and l10n_ch_zip

BT-aestebanez:


leemannd:
LGTM, but should we deactivate test on DTA while it's unported? cc @yvaucher

l10n_ch_states/__manifest__.py
```diff
@@ -22,15 +22,19 @@
     'name': 'Switzerland Country States',
     'category': 'Localisation',
     'summary': '',
-    'version': '9.0.1.0.0',
-    'author': 'copado MEDIA UG, Odoo Community Association (OCA)',
+    'version': '10.0.1.0.0',
+    'author': '''
+        copado MEDIA UG,
+        brain-tec AG,
```
yvaucher, line 28: I don't see a significant enough addition to the module to set brain-tec as author here. Please remove this line.
While you contribute, you are not an author of the module.

yvaucher:
This PR is redundant with #287

l10n_ch_states/__manifest__.py
BT-aestebanez: Yes, you are right

BT-aestebanez:
@yvaucher Ok, so I think the best idea would be apply these small changes in the PR 287 and close this one, do you agree?

yvaucher:
I see that there are a few missing changes both in this PR and in #287, I'll do the portage for both modules. I want to clean few things like copyright headers and readme.
Plus I'll make a PR per module.

BT-aestebanez:
Ok, thanks

yvaucher:
superseeded by #292 and #293

PR #294, opened 2017-01-03
Wrong method signature

stephanamann:
Method of base class has no data parameter! Creating DTA failed with this module installed.


yvaucher:
This was fixed here

https://github.com/OCA/l10n-switzerland/commit/8b30f787d3dabb9072ee0517cf0e5e92daa1038f

PR #298, opened 2017-02-07
Allows you to edit DTA code on partner banks

open-net-sarl:


l10n_ch_dta/bank_view.xml
```diff
@@ -6,11 +6,9 @@
       <field name="model">res.partner.bank</field>
       <field name="inherit_id" ref="base.view_partner_bank_form"/>
       <field name="arch" type="xml">
-        <group name="bank" position="after">
-          <group string="DTA options" attrs="{'invisible': [('company_id', '=', False)]}">
```
yvaucher, line 10: company_id = False meant that we wanted to display the config only for the bank of the company and not on the bank of the suppliers or customers.

Is there no way to reproduce this ?

l10n_ch_dta/__openerp__.py
```diff
@@ -30,9 +30,8 @@
              'account_payment_order',
              'document'],
  'data': ['data/account_payment_method.xml',
-          ],
- 'demo': ['../account/test/account_minimal_test.xml',
```
yvaucher, line 34: This was fixed in https://github.com/OCA/l10n-switzerland/pull/299/commits/cd2258ad176bd6554dc3ad1db677d9ae6a8649b2

yvaucher:
You need to rebase this PR.

Could you also explain why we need to be able to edit the `dta_code`

open-net-sarl:
This PR is too old (useless). I close it.

PR #300, opened 2017-02-08
[8.0] FIX l10n_ch_payment_slip: when country is not present in 'Address Format' , do not remove part of the address.

eLBati:
To reproduce the problem: customer has country_id but country_id.address_format does not contain 'country_name' nor 'country_id'.
Printing payment slip, line '%(city)s %(state_code)s %(zip)s' is not diplayed.

### DIFF
--- a/l10n_ch_payment_slip/payment_slip.py
+++ b/l10n_ch_payment_slip/payment_slip.py
@@ -483,9 +483,6 @@ class PaymentSlip(models.Model):
         text.moveCursor(0.0, font.size)
 
         address_lines = com_partner.contact_address.split("\n")
-        if com_partner.country_id:
-            del address_lines[-1]
-
         for line in address_lines:
             if not line:
                 continue


l10n_ch_payment_slip/payment_slip.py
```diff
@@ -483,9 +483,6 @@ def _draw_address(self, canvas, print_settings, initial_position, font,
         text.moveCursor(0.0, font.size)
 
         address_lines = com_partner.contact_address.split("\n")
-        if com_partner.country_id:
```
yvaucher, line 486: This will display the country when we don't want it on the BVR.

For your specific case, you should at least test if country is part of self.country_id.address_format.
Or to make sure it works in any case we should rely on our own address format.


Maybe in an onchange context we could set the com_partner.address_format temporarily.
```
with self.env.do_in_onchange():
    com_partner.country_id.address_format =  "%(street)s\n%(street2)s\n%(city)s %(state_code)s %(zip)s" # or the right format without country_name
    address_lines = com_partner.contact_address.split("\n")
```

yvaucher (changes requested):

l10n_ch_payment_slip/payment_slip.py
eLBati: > For your specific case, you should at least test if country is part of self.country_id.address_format

Hi @yvaucher , in the first place, I made the change as you suggest, with something like:

```
if 'country' in com_partner.country_id.address_format:
    del address_lines[-1]
```
then I looked at version 9:
https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_payment_slip/models/payment_slip.py#L471
and made this PR, assuming that, if users don't want to print the country , they just remove country from address_format.

That being said, which solution do you suggest?
Thanks

yvaucher: @eLBati version 8.0 is quite old now, so I would prefer not to break something working.

So it would be better with backward compatibility.

This fix seems to have been applied only in 8.0 and didn't made it in 9.0
https://github.com/OCA/l10n-switzerland/commit/27d8fa88152927d21bf6cd14be709b15034e21f0

as it landed after the portage to 9.0

@johannbelet Any thought about this ?

yvaucher: https://github.com/OCA/l10n-switzerland/pull/263

johannbelet: Some people need to keep contry in adress format for other purposes but country is usually not needed on BVR

yvaucher: Here is a proposal for 9.0 https://github.com/OCA/l10n-switzerland/pull/303 and future versions
A bit more complicated but should cover all the needs and by default will work without further configuration.

PR #309, opened 2017-03-10
Fixed length issue 

ghost:
Fixed length issue in ESR fields, max 23 characters of name and address are take to not overlap with layout.

### DIFF
--- a/l10n_ch_payment_slip/models/payment_slip.py
+++ b/l10n_ch_payment_slip/models/payment_slip.py
@@ -460,18 +460,29 @@ class PaymentSlip(models.Model):
         :type com_partner: :py:class:`openerp.models.Model`
 
         """
+        longest_line = None
+
+        if font.size >= 11:
+            longest_line = 23
+        elif font.size == 10:
+            longest_line = 27
+        elif font.size == 9:
+            longest_line = 30
+        elif font.size <= 8:
+            longest_line = 34
+
         x, y = initial_position
         x += print_settings.bvr_add_horz * inch
         y += print_settings.bvr_add_vert * inch
         text = canvas.beginText()
         text.setTextOrigin(x, y)
         text.setFont(font.name, font.size)
-        text.textOut(com_partner.name[:23])
+        text.textOut(com_partner.name[:longest_line])
         text.moveCursor(0.0, font.size)
         for line in com_partner.contact_address.split("\n"):
             if not line:
                 continue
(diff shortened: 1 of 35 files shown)


BT-astauder:
👍 +1

open-net-sarl:
👍 +1

l10n_ch_payment_slip/models/payment_slip.py
```diff
@@ -466,12 +466,12 @@ def _draw_address(self, canvas, print_settings, initial_position, font,
         text = canvas.beginText()
         text.setTextOrigin(x, y)
         text.setFont(font.name, font.size)
-        text.textOut(com_partner.name)
+        text.textOut(com_partner.name[:23])
```
yvaucher: I'm afraid in that case that we would loose valuable informations.
Let say I have a long street name, I could loose the street number.

Plus, with a different font size of 9 you can put 30 characters

Maybe we could make add an option to compute the font size depending on number of characters of the longest line of the address.

It would be:
longest_line <= 23 = size 11
longest_line <= 27 = size 10
longest_line <= 30 = size 9
longest_line <= 34 = size 8

according to my tests.

yvaucher (changes requested):

yvaucher (approved):

yvaucher:
Thanks for the improvement, could you also solve the conflict by doing a rebase ?

l10n_ch_payment_slip/models/payment_slip.py
```diff
@@ -481,16 +481,40 @@ def _draw_address(self, canvas, print_settings, initial_position, font,
...
         text = canvas.beginText()
         text.setTextOrigin(x, y)
-        text.setFont(font.name, font.size)
-        text.textOut(com_partner.name)
+        text.setFont(font.name, font_size)
+        text.textOut(com_partner.name[:cutoff_length])
+        # we are moving in the original font size to new position
         text.moveCursor(0.0, font.size)
-        [text.textLine(l) for l in self._get_address_lines(com_partner) if l]
+        [text.textLine(l[:cutoff_length])
```
yvaucher: l10n_ch_payment_slip/models/payment_slip.py:515:42: W291 trailing whitespace

yvaucher:
Squashed commit and changed commit message at https://github.com/OCA/l10n-switzerland/commit/e0b57d53cce6c15f5403fa05b0d07888ee172760

PR #312, opened 2017-03-15
Allow whitespaces in search of invoice references

guewen:
When searching in Vendor references, we want to ignore the whitespaces.
This is because very often, it will contain BVR/ESR number that contains
whitespaces. And the user usually searches for the invoice number which
is only a part of the BVR/ESR number.

Example:
* the vendor reference on the supplier invoice is: 27 29990 00000 00001 70400 25019
* the user knows the invoice's supplier reference: 17040025
* when searching 17040025, he expects to find the invoice
* the reverse situation, where the ref contains no whitespaces and the
  user searches for a string with whitespaces might also happen

### DIFF
--- a/l10n_ch_base_bank/models/invoice.py
+++ b/l10n_ch_base_bank/models/invoice.py
@@ -23,13 +23,21 @@ class AccountInvoice(models.Model):
                 domain.append(arg)
                 continue
             if operator not in ('like', 'ilike', '=like', '=ilike',
-                                'not like', 'not ilike', '=',
-                                '!=', '<', '<=', '>', '>=', 'in', 'not in'):
+                                'not like', 'not ilike'):
                 domain.append(arg)
                 continue
             if value:
                 value = value.replace(' ', '')
-                if operator in ('like', 'ilike', 'not like', 'not ilike'):
+                if not value:
+                    # original value contains only spaces, the query
+                    # would return all rows, so avoid a costly search
+                    # and drop the domain triplet
+                    continue
+                # add wildcards for the like search, except if the operator
+                # is =like of =ilike because they are supposed to be there yet
+                if operator.startswith('='):
+                    operator = operator[1:]
+                else:
                     value = '%%%s%%' % (value,)
             query = ("SELECT id FROM account_invoice "
                      "WHERE REPLACE(reference, ' ', '') %s %%s" %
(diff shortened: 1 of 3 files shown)


simahawk (approved):

yvaucher (approved):
Nice change

l10n_ch_base_bank/models/invoice.py
```diff
@@ -11,6 +11,37 @@ class AccountInvoice(models.Model):
...
+            if value:
+                value = value.replace(' ', '')
+                if operator in ('like', 'ilike', 'not like', 'not ilike'):
+                    value = '%%%s%%' % (value,)
+            query = ("SELECT id FROM account_invoice "
+                     "WHERE REPLACE(reference, ' ', '') %s %%s" %
+                     (operator,))
+            self.env.cr.execute(query, (value,))
+            ids = [t[0] for t in self.env.cr.fetchall()]
+            domain.append(('id', 'in', ids))
```
gurneyalex, line 47: Don't you need to special if `ids == []`?

l10n_ch_base_bank/models/invoice.py
```diff
@@ -11,6 +11,37 @@ class AccountInvoice(models.Model):
...
+            if field != 'reference':
+                domain.append(arg)
+                continue
+            if operator not in ('like', 'ilike', '=like', '=ilike',
+                                'not like', 'not ilike', '=',
+                                '!=', '<', '<=', '>', '>=', 'in', 'not in'):
+                domain.append(arg)
+                continue
+            if value:
+                value = value.replace(' ', '')
```
gurneyalex: I think you need to handle the special case where value.replace(' ', '') -> empty string. This will match *all* invoices and kill your perfs, and it may happen with incremental searches.

gurneyalex (commented):
a few questions asked.

l10n_ch_base_bank/models/invoice.py
guewen: if `ids == []`, the ORM automatically translates it to:

```
SELECT "account_invoice".id FROM "account_invoice" WHERE FALSE ORDER BY "account_invoice"."date_invoice" DESC,"account_invoice"."number" DESC,"account_invoice"."id" DESC
```

guewen:
Added a new commit that remove the path for the operators which are not in the LIKE family, they will now be treated as usual. Only the LIKE (and family) operators will now search without whitespaces. The commit includes some details.

gurneyalex (approved):

PR #322, opened 2017-03-30
Add a generic BVR/ESR reference matching module

ghost:
Trivial reinject of the V11 importer's hardcoded search over move lines.  The original is at https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_payment_slip/wizard/bvr_import.py#L160. 

This eases the transition from V11 to CAMT at least for us and our clients, so we're sharing it.  Maybe there's a better option somewhere in the account-bank-statement machinery, but AFAIK this is the simplest way to do the job.  It's also true that the code isn't specifically Swiss, so maybe this could be moved somewhere else?


ghost:
Looks like this is useless on a standard Odoo setup because the reconciliation widget is smart enough to do exactly what this module does.  I created this module based on a very old version of the reconciliation widget, (too long to explain) but it's useless on current 9.0 so I'm sorry, forget about it.

PR #326, opened 2017-04-20
[FIX][9.0] avoid crash if com_partner has no name

ghost:
This bit a client today.

Introduced by e0b57d53cce6c15f5403fa05b0d07888ee172760; since it has been forward ported in 10.0 by #325, this should be as well.


yvaucher (approved):
Indeed name is not mandatory anymore.

tafaRU (approved):

PR #329, opened 2017-05-03
Update res.better.zip.csv

leNeo:
suppression de l'indication du canton dans la localité


yvaucher:
Why do you want to remove this ?

This is a valuable information for the user as you don't need to know by heart the zip code to know if you are talking for instance about Cressier FR or Cressier NE

yvaucher:
Okay we have this information already in the display name:

https://github.com/OCA/partner-contact/blob/10.0/base_location/models/better_zip.py#L39

yvaucher (approved):

PR #331, opened 2017-05-10
[10.0] l10n_ch_bank - Remove dependency on l10_ch

### LINKED ISSUE #327: remove l10n_ch chart as dependence of l10n_ch_bank
l10n_ch_bank should not depend on l10n_ch as there is no link between bank data and chart of account.
That might lead to conflicts if company has its own chart of accounts.

yvaucher:
Fixes #327


Martronic-SA:
Many thanks!
Could you do the same for v 9.0 ?

yvaucher:
@Martronic-SA done in #332

PR #336, opened 2017-06-12
Contraint on BVR adherent number / CPP pair

Martronic-SA:
First of all there is a confusion for the BVR coding line:
0100000600007>103559000000000020170013270+ 100007254>
.....................................^ reference beginning with......................^ adherent number
.....................................bank client account number
.....................................( 6 chars )
You are naming the 6 chars client account number as BVR adherent number which is false.
Here Raiffeisen documentation : [Raiffeisen doc pdf](https://www.raiffeisen.ch/content/dam/www/rch/pdf/firmenkunden/zahlungsverkehr-und-konten/fr/fiche-produit-bulletin-versement-orange.pdf)
However, both of these numbers can be used many times but the pair (adherent number / client account number) must be unique.

### DIFF
--- a/l10n_ch_base_bank/models/bank.py
+++ b/l10n_ch_base_bank/models/bank.py
@@ -349,5 +349,5 @@ class ResPartnerBank(models.Model, BankCommon):
             if 'Bank/CCP' in self.acc_number:
                 self.acc_number = self._get_acc_name()
 
-    _sql_constraints = [('bvr_adherent_uniq', 'unique (bvr_adherent_num)',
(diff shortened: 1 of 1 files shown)


yvaucher (approved):
Indeed in case of BVRB the scan line has first the customer identification number and at the end the adherent number of the bank (which is it's ccp).

This fix is fine as it makes sens that each bank manage their own customer id registering.

yvaucher:
For the mix between adherent number and customer id ref, we should rename the field in next version.

For now I will merge this fix and report it in 10.0.

PR #337, opened 2017-06-15
10.0 migrating l10n ch scan bvr to 10.0

BT-aestebanez:


BT-jmichaud:
@yvaucher :
Hello,
I saw you did the review for the migration from 8.0 to 9.0.
Should it still be done by yourself or should I contact anyone else?
Thank you.
Regards,

Michaud Joel  | Projektleiter & Software Engineer
BSc Wirtschaftsinformatik FH 
brain-tec AG | Saflischstrasse 4, CH-3900 Brig
Tel: +41 (0)44 552 01 20 | Dir: +41 (0)27 552 02 14
www.braintec-group.com | joel.michaud@braintec-group.com

yvaucher:
@BT-jmichaud anyone can do review. Here are OCA rules:

https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#review

BT-jmichaud:
This was reviewed and tested on our side and it works. OK for us.

ghost:
The travis failure is unrelated to the changes. (#339)

ghost (approved):
Thank you for your work. We conducted functional tests on this PR and they were all successful. We have been using this module in previous versions and we felt at home with this version.

l10n_ch_scan_bvr/wizard/scan_bvr.py
```diff
@@ -42,6 +29,19 @@ def _default_journal(self):
         else:
             return False
 
+    journal_id = fields.Many2one(comodel_name='account.journal',
+                                 string="Invoice journal",
+                                 default=lambda self: self._default_journal())
```
yvaucher: What is the reason for this change ?

l10n_ch_scan_bvr/wizard/scan_bvr.py
```diff
@@ -158,6 +158,9 @@ def _create_invoice_line(self, data):
             specs = invoice_line_model._onchange_spec()
             default_values = invoice_line_model.default_get(specs)
             invoice_line_vals = specs.copy()
+            for val in invoice_line_vals:
+                if invoice_line_vals[val]:
+                    invoice_line_vals[val] = None
```
yvaucher: I don't get why you empty all values of this dictionary.

If you need an empty dict you can do:

``` python
invoice_line_vals = {k: None for k in specs.keys()}
```

yvaucher (changes requested):
A couple of questions

l10n_ch_scan_bvr/wizard/scan_bvr.py
BT-aestebanez: I tried to follow OCA guidelines (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models)
Specifically:
In a Model attribute order should be:
1 Private attributes (_name, _description, _inherit, ...)
2 Default method and _default_get   <--------------------
3 Fields declarations

And https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#fields
* Default functions should be declared with a lambda call on self. The reason for this is so a default function can be inherited. Assigning a function pointer directly to the default parameter does not allow for inheritance.

l10n_ch_scan_bvr/wizard/scan_bvr.py
BT-aestebanez: You are right, thanks for the comment.
For the record, I need an empty dict because otherwise the values are not modified correctly in the onchange(): Some values are not updated there and receive the value '1', which is not correct.

BT-aestebanez:
@yvaucher Thanks for the review
I have updated the code with your suggestion, and explained the other comment

BT-aestebanez:
Should I modify something else?
To the best of my knowledge the checks are not related to my code

yvaucher (approved):

yvaucher:
Ok let's merge this. The failing checks are on code coverage. 

l10n_ch_scan_bvr/wizard/scan_bvr.py deserves more testing 62 lines are untested.

PR #340, opened 2017-06-21
[9.0] DTA: Use aggregated bank_lines instead of payment_lines

BT-astauder:
Currently the DTA is created based on the payment_lines that ahave a 1:1
 relation to account_move_lines. Especially to pay expenses this creates
 a transaction for each expense.
 bank_lines (already existing) have merged all payments to the same
 partner when there is no structured communication used. It is better to
 use bank_lines.

 Payment to the same partner but with different BVR are
 not merged!


BT-astauder:
@yvaucher May you please have a look into that? Thank you

BT-astauder:
See patch for 10.0 in #341

yvaucher (approved):

yvaucher:
Thanks

PR #342, opened 2017-07-05
[MIG][10.0] Postfinance modules

ecino:
- l10n_ch_lsv_dd
- l10n_ch_account_statement_import_postfinance
- l10n_ch_fds_postfinance
- l10n_ch_fds_postfinance_upload_dd
- l10n_ch_fds_postfinance_upload_sepa

- [ ] Depends on new camt_parser https://github.com/OCA/bank-statement-import/pull/119


ghost (commented):
The `l10n_ch_fds_postfinance/` section of this PR LGTM and we're currently using it in production. I haven't used the other modules nor reviewed the changes.

~~One little thing I'd like to be modified is that there's just too many commits, most of which are already part of the 9.0 branches (or rebases of such) (unless I'm mistaken). it would be more helpful for the reviewers to have, say, one commit that copies l10n_ch_fds_postfinance from its state in, for instance, e7f9446d3d54cf10671dab644ee705532ed8887e as is, and then one or more other commits that bump the version, rename the manifest, fix the imports, and everything.~~

ecino:
Last time I had not the full history of version 9.0 I was asked to include that. So I don't know what's best.

ghost:
OK then, it's my habits vs the official procedure, so forget it and follow the official procedure.

ghost:
Travis complains that `exportKey` doesn't take a config argument. Does not occur in production. @ecino mind having a look at it? Did PyCrypto break its API at some point?

ecino:
It's weird, by looking at the API it seems the method never had a config parameter. I will check and test that.

ecino:
I will do separate pull requests for the different modules to help the review process.

PR #346, opened 2017-07-20
[8.0][FIX] view inheritance on res.partner

grindtildeath:
Field credit_limit is not defined on base.view_partner_form


yvaucher (approved):
Checked on runbot, pep8 error raised by travis are unrelated. As it break a view I fast-track it.

PR #351, opened 2017-08-24
[FIX] write BVR adherent number on bank account when the user select …

vrenaville:
When you scan an new supplier invoice, you need to
create the account but you don't know the bvr adherent number (is it stored in the BVR scan line).

So you scan the BVR Scan line with you paypen in the scan BVR view, it will search for a bank account with the bvr adhrerent number reference stored in the scan line
but it will be not found , so you select the partner and the account manually.

currently you need to find yourself the bvr adherent number in the scan line report it on the partner bank account, to be able to find find it directly.

with this PR you just need to select the account the first time, it will copy the bvr adherent number found in the scan line into the related bank account.

The next time the correct bank account will be found directly.


yvaucher (approved):
This was removed from

https://github.com/OCA/l10n-switzerland/commit/3967bc1b5f9decb0734f9a4bd3a51a3bc5738325

PR #356, opened 2017-10-04
[9.0] Fix import of uncompressed PostFinance camt files

MaxyMoos:
Import of uncompressed PostFinance-generated camt files currently fails, as in this scenario, the parser's `self.attachments` variable is never initialized but is still accessed in the `parse` method, leading to an AttributeError being raised.

This simple fix works because the `_get_attachments_from_stream` method correctly handles untarred files - by setting `self.attachments` to an empty dict.

### DIFF
--- a/l10n_ch_bank_statement_import_postfinance/models/postfinance_file_parser.py
+++ b/l10n_ch_bank_statement_import_postfinance/models/postfinance_file_parser.py
@@ -83,8 +83,7 @@ class XMLPFParser(models.AbstractModel):
         """
         self.tar_source = data_file
         self.data_file = self._get_content_from_stream()
-        if self.is_tar:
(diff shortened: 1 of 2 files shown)


MaxyMoos:
Ok seems like the test case `test_attachments_detection` expects `self.attachments` not to be set to an empty dict when no attachments are found, but explicitly untouched.

The test can easily be adapted to suit the change of behavior I implemented, but I'll wait for your go first.

MaxyMoos:
I took the liberty of updating the test that was previously failing.
Now the build fails for what looks like something unrelated (missing CHF currency in the test database from what I get from the logs).

yvaucher:
@ecino could you review this? Seems ok to me.

ecino:
LGTM :+1:

PR #359, opened 2017-10-27
Porting l10n_ch_payment_slip_layouts version 8.0 to 10.0 of the module

dynapps-ch-jean:
This version is now compatible with version 10.0 of Odoo.
I added as well a shortcut button to reach the invoice with BVR slip directly from an invoice. We can even use that file to "Send by email".

Jean


l10n_ch_payment_slip_layouts/__manifest__.py
```diff
@@ -19,7 +19,7 @@
 #
 ##############################################################################
 {'name': 'Payment slip alternate layout(s)',
- 'version': '8.0.0.1.0',
+ 'version': '10.0.0.1.0',
```
leemannd: s/10.0.0.1.0/10.0.1.0.0

l10n_ch_payment_slip_layouts/models/invoice.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# © 2012-2016 Camptocamp SA
```
leemannd: s/©/Copyright

l10n_ch_payment_slip_layouts/view/account_invoice.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0"?>
+<openerp>
+  <data>
```
leemannd: `s/openerp/odoo`  & `<data>` can be removed

leemannd (changes requested):
Hello, 
welcome and thanks for the port!
You'll find nitpicking to fit OCA standards.

Also could you update python files headers with the following template? => https://github.com/OCA/maintainer-tools/blob/master/template/module/models/model_name.py#L3

dynapps-ch-jean:
Hi,

Thanks for your post. I just did update as required. Is it fine ?

Best regards,

l10n_ch_payment_slip_layouts/__init__.py
```diff
@@ -1,23 +1,7 @@
...
-#    This program is distributed in the hope that it will be useful,
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2017 Jean Respen and Nicolas Bessi
```
leemannd: Quite good. It should be:
Copyright 2015 Camptocamp SA -  Nicolas Bessi
Copyright 2017 Jean Respen

dynapps-ch-jean: Thanks, it's done !

dynapps-ch-jean:
Hi,

As I am fairly new to Odoo module development, I still have an error to fix, which is in the unity tests.

It seems that this line 
gen = self.registry['report']._compute_documents_list(
                [self.invoice.id],
                context={}
            )

is incorrect, due to ` TypeError: unbound method _compute_documents_list() must be called with report instance as first argument (got list instance instead)

Do you have any idea ?

Thanks

dynapps-ch-jean:
Hi,

Okay I found the bug, travis passed.

Do I need to pass the three remaining tests as well to get my pull request accepted ?

Best,

README.md
```diff
@@ -94,6 +94,7 @@ addon | version | summary
 [l10n_ch_scan_bvr](l10n_ch_scan_bvr/) | 10.0.1.0.0 | Switzerland - Scan ESR/BVR to create invoices
 [l10n_ch_states](l10n_ch_states/) | 10.0.1.0.0 | Switzerland Country States
 [l10n_ch_zip](l10n_ch_zip/) | 10.0.1.0.1 | Provides all Swiss postal codes for auto-completion
+[l10n_ch_payment_slip_layouts](l10n_ch_payment_slip_layouts/) | 10.0.1.0.0 | Add new BVR/ESR payment slip layouts like invoice with slip on same document
```
yvaucher, line 97: You don't need to make that edition, we have a bot that take care of it. (no need to change, just an advise for next time)

l10n_ch_payment_slip_layouts/__init__.py
```diff
@@ -1,23 +1,8 @@
...
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2015 Camptocamp SA - Nicolas Bessi
+# Copyright 2017 Jean Respen
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
 from . import report
 from . import payment_slip
 from . import company
```
yvaucher, line 7: `company` and `payment_slip` files should be moved in `models` folder

l10n_ch_payment_slip_layouts/report/__init__.py
```diff
@@ -1,21 +1,5 @@
...
-#    but WITHOUT ANY WARRANTY; without even the implied warranty of
-#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-#    GNU Affero General Public License for more details.
-#
-#    You should have received a copy of the GNU Affero General Public License
-#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
-#
-##############################################################################
+# Copyright 2015 Camptocamp SA - Nicolas Bessi
+# Copyright 2017 Jean Respen
```
yvaucher, line 3: Please don't add your copyright when you don't make code change.

l10n_ch_payment_slip_layouts/report/payment_slip_from_invoice.py
```diff
@@ -58,41 +38,36 @@ def _compute_documents_list(self, cr, uid, invoice_ids,
...
+        docs = self._compute_documents_list(ids,
                                             report_name=report_name,
                                             context=context)
         if company.merge_mode == 'in_memory':
             return self.merge_pdf_in_memory(docs)
         return self.merge_pdf_on_disk(docs)
 
     @api.v7
-    def get_pdf(self, cr, uid, ids, report_name, html=None, data=None,
+    def get_pdf(self, ids, report_name, html=None, data=None,
```
yvaucher, line 59: In Odoo 10.0 old API compatibility was dropped, I think you can remove the @api.v7 decorator

And remove context from the signature

https://github.com/odoo/odoo/blob/5d0d80afa5bce8780305173c93e291466a133f7f/addons/report/models/report.py#L146

l10n_ch_payment_slip_layouts/report/payment_slip_from_invoice.py
```diff
@@ -58,41 +38,36 @@ def _compute_documents_list(self, cr, uid, invoice_ids,
 
     @api.v7
     def _generate_inv_and_one_slip_per_page_from_invoice_pdf(
-            self, cr, uid, ids, report_name=None, context=None):
+            self, ids, report_name=None, context=None):
```
yvaucher, line 41: You can remove context from signature as you don't use it in `compute_documents_lists()` and it should already available through `self.env.context`.

l10n_ch_payment_slip_layouts/report/payment_slip_from_invoice.py
```diff
@@ -58,41 +38,36 @@ def _compute_documents_list(self, cr, uid, invoice_ids,
...
         PDF are merged in memory or on
         file system based on company settings
 
         :return: the generated PDF content
         """
-        user_model = self.pool['res.users']
-        company = user_model.browse(cr, uid, uid, context=context).company_id
-        docs = self._compute_documents_list(cr, uid, ids,
+        user_model = self.env['res.users']
+        company = user_model.browse(self.env.uid).company_id
```
yvaucher, line 50: Here you can simply do:

``` python
company = self.env.user.company_id
```

l10n_ch_payment_slip_layouts/tests/test_layout.py
```diff
@@ -119,22 +119,18 @@ def setUp(self):
...
-                self.env.cr,
-                self.env.uid,
+            gen = self.env['report']._compute_documents_list(
                 [self.invoice.id],
                 context={}
             )
             self.assertIsInstance(gen, types.GeneratorType)
             pdfs = [x for x in gen]
-            self.assertTrue(len(pdfs) == 2)
+            self.assertTrue(len(pdfs) == 1)
```
yvaucher, line 128: Not so sure of that change, do you know why we get only 1 pdf instead of 2 ?

yvaucher (changes requested):
Hello, thanks for your first contribution !

Please see my inline comment.

I'm a bit concerned about the change in the tests from 2 pdf to 1. Can you explain it to me ?

yvaucher:
For the code coverage it would be great to have some more tests.

According to coverage we don't test what is in payment_slip_from_invoice.py which is strange to me. We would probably need to mock the merge_pdf_in_memory and merge_pdf_on_disk.

But we probably don't want to block this PR on that.

PR #371, opened 2017-11-17
[MIG] l10n_ch_base_bank: Migration to 11.0

yvaucher:
Remove coding UTF-8 from headers
Reset version to 11.0.1.0.0
Update README.rst


yvaucher:
sql injection is a false positive, as operators are filtered with a continue in the for loop.

PR #374, opened 2017-11-22
[10.0][FIX] write BVR adherent number on bank account when the user select …

grindtildeath:
…the correct bank in the wizard for the next search match directly with the bvr adherent number

Forward port of https://github.com/OCA/l10n-switzerland/pull/351


jcoux (approved):
Code seems good :+1: 
No tests.

yvaucher (approved):

PR #375, opened 2017-11-27
[IMP][TRANSLATIONS] Updating german translations of l10n_ch_scan_bvr

BT-aestebanez:


BT-aestebanez:
Some translations are not working. See f.e.: https://github.com/OCA/l10n-switzerland/blob/10.0/l10n_ch_scan_bvr/i18n/de.po#L93

yvaucher (approved):
Checked on runbot and the translation are there in german

PR #381, opened 2017-12-04
[UPDATE] Changed Non-ASCII character

magneticlab-ch:
Error happening while installing as submodule on odoo.sh
l10n_ch_zip/models/better_zip.py on line 39, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details


yvaucher:
@magneticlab-ch You should run your odoo 11.0 with **python3**, encoding are set by default to UTF-8.
There is no need for a encoding declaration anymore.

Odoo 11.0 do not officially supports python 2.7

Thus I close this PR

PR #382, opened 2017-12-04
Add sudo() when creating a temp country

mdietrichc2c:
Small PR to correct an issue : if a user with no creationrights on the "res.country" object prints a BVR, the zip and city will be inverted, as the default address format will be used.


ghost (approved):
LGTM!

gvellut-c2c (approved):

yvaucher (approved):

PR #383, opened 2017-12-08
[10.0] Fix UserError falling when trying to read unexisting field

### LINKED ISSUE #378: V10: can't validate a payment order if a CCP number is required but missing on the res.partner.bank
Hi

The problem is that in such a case, the message that should have popped is trying to get the value of the "name" field of that res.partner.bank object.
Located here:
l10n_ch_pain_base/models/account_payment_order.py, line 94
Whereas at line 77 it's correct.

yvaucher:
Fixes #378


yvaucher:
@open-net-sarl could you review this one as you spotted it

PR #385, opened 2017-12-18
[10.0]l10n_ch_scan_bvr : Move field supplier_invoice_default_product to purchase group in view

grindtildeath:


yvaucher (approved):
This is a trivial one I will fast track it

PR #387, opened 2017-12-20
Rename BVR and BVR/ESR to ISR in english and make it the base term

yvaucher:
Renaming and migration script for `l10n_ch_base_bank` module

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Provides the list of Swiss postal ZIPs for auto-completion.
 l10n_ch_payment_slip
 --------------------
 
-Adds ESR/BVR report on invoice. Every ESR/BVR element position can be configured independently by company.
+Adds ISR (PVR/BVR/ESR) report on invoice. Every ISR element position can be configured independently by company.
 Multiple payment terms on invoices are supported.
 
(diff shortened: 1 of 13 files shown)


davidwul:
:+1: 
Make sense. Why not rename also CCP ( french for Compte de Chèque Postaux) by CP (Compte Postal) in French and PC (Postkonto) in german... but no idea in english ... I would put PC

yvaucher:
That's a good idea, I didn't check what it would require to change CCP.

On the other hand I find CCP more accurate than CP and PC abreviations can be ton of things.
But I'm biased as my mother tongue is french.

By the way isn't it PC-Konto or CP-Konto instead of just PC ?
In the de.po files we currently have CP-Konto but that could be wrong.

I could put postal number/account in english to be translated by CCP and PC-Konto

yvaucher:
Seems the CP-Konto in translations is a mistake

davidwul:
Yes PC-Konto is clearer than just PC.
And for me postal account is good in english.

yvaucher:
I'll merge this renaming as stand alone, renaming CCP deserve an other PR

Issue #390, opened 2018-01-26
CH Payroll: update parameter default values in the payroll configuration


treviser:
Just for the convenience: please update the following parameter defaults according to the new legal values:

- OAI/II/IC: 5.125%
- OBP minimum legal: 1,767.633
- UI Maximum limit: 12350

Thanks!

didierdonze:
- OAI/II/IC: 5.125%
-> I agree
- OBP minimum legal: 1,767.633
-> The annual minimal salary amount to contribute is 21150 (monthly 1762.5)
- UI Maximum limit: 12350
-> I agree

treviser:
> OBP minimum legal: 1,767.633
> -> The annual minimal salary amount to contribute is 21150 (monthly 1762.5)

I agree, stupid me :-)

asaunier:
@treviser Feel free to have a look at PR #394 for Odoo 10.

Are the changes required for Odoo 9 as well?

It seems some values had already been updated.
For instance https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_hr_payroll/data/hr.salary.rule.xml#L108 (UI maximum limit)

On the other hand the LPP min seems to be outdated:
https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_hr_payroll/data/hr.salary.rule.xml#L211
Ditto for OAI:
https://github.com/OCA/l10n-switzerland/blob/9.0/l10n_ch_hr_payroll/data/hr.salary.rule.xml#L70

Proposed changes for version 9: https://github.com/OCA/l10n-switzerland/pull/395

asaunier:
PRs for both Odoo 9 and 10 have been merged.

PR #394, opened 2018-02-08
[10.0] l10n_ch_hr_payroll: update some default values

asaunier:
Fixes https://github.com/OCA/l10n-switzerland/issues/390

I have only changed OBP minimum legal (``lpp_min``) and UI Maximum limit (``ac_limit``).
OAI/II/IC rate was already at 5.125%

The second commit edc7d6e is not really mandatory: tests pass without it. But it makes sense to update the example data with the new ``ac_limit`` and ``lpp_min`` default values.
But then the ``wage_fulltime`` of the example is no longer above the UI maximum limit => we had to upgrade this wage and the related tested values.

### DIFF
--- a/l10n_ch_hr_payroll/models/res_company.py
+++ b/l10n_ch_hr_payroll/models/res_company.py
@@ -36,7 +36,7 @@ class ResCompany(models.Model):
     # UI(AC)
     ac_limit = fields.Float(
         string='Maximum limit',
-        default='10500',
+        default=12350,
         digits=dp.get_precision('Account'),
         required=False)
     ac_per_off_limit = fields.Float(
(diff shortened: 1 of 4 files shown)


yvaucher (approved):

mpanarin (approved):
other than, imo, not the best commit message - LGTM
https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#commit-message

asaunier:
@yvaucher @mpanarin Thanks for reviewing!

I have squashed the 2 original commits into a single one to make the PR simpler. I had initially made 2 distinct commits because the changes in ``test_max_lpp.py`` were quite cumbersome because of many hard-coded tested values I had to update.

I have also changed the PR and commit messages in order to follow the guidelines, even though the commit message is a bit longer than 50 chars but it seems reasonable to me.

asaunier:
The suggested changes cannot be cherry-picked to Odoo 9 because the addon has quite changed from 9 to 10. Here are proposed changes for version 9: https://github.com/OCA/l10n-switzerland/pull/395

PR #398, opened 2018-02-24
[IMP] l10n_ch_bank: bic and city required on view for switzerland

mpanarin:
updated csv by filling the country column and made fields bic and city required when Switzerland is selected for a bank, as sometimes if the user forgets these 2 fields, the bank can refuse the payment.


yvaucher (approved):

yvaucher:
travis errors are not related

PR #400, opened 2018-03-03
[IMP] l10n_ch_bank: bic and city required on view for switzerland

mpanarin:
updated csv by filling the country column and made fields bic and city required when Switzerland is selected for a bank, as sometimes if the user forgets these 2 fields, the bank can refuse the payment.


l10n_ch_bank/views/res_bank.xml
```diff
@@ -0,0 +1,25 @@
...
+            </xpath>
+
+            <xpath expr="//field[@name='bic']" position="attributes">
+                <attribute name="attrs">{'required': [('country_code', '=', 'CH')]}</attribute>
+            </xpath>
+
+        </field>
+    </record>
+
+</odoo>
```
yvaucher, line 25: I think this should go in the `l10n_ch_base_bank` module.

The current module `l10n_ch_bank` is here for the bank data while the module `l10n_ch_base_bank` is about the changes made on the bank object.

yvaucher (changes requested):

asaunier:
Should we close this PR, since it is superseeded by #406?

PR #402, opened 2018-03-07
[ADD] l10n_ch_import_isr_v11: new module

mpanarin:
Due to changes in L10n_ch it is partly supporting ISR it was decided to move this wizard from l10n_ch_payment_slip into a separate module
@yvaucher 

Depends on:

- [x] ~~https://github.com/OCA/account-reconcile/pull/190~~
- [x] https://github.com/OCA/account-reconcile/pull/206


open-net-sarl:
V11 will disappear on June 30th, cf. https://www.paymentstandards.ch/en/home/splash.html...

yvaucher:
@open-net-sarl yep but can still be needed it since then. We can drop it afterward.

Having it out of l10n_ch_payment_slip allows to use the qweb version of ISR of Odoo core.

yvaucher:
@mpanarin dependency was merged can you drop your last commit ?

PR #403, opened 2018-03-28
[10.0] l10n_ch_payment_slip: fix `_get_address_lines`

simahawk:
l10n_ch_payment_slip: fix `_get_address_lines`
    
    `payment_slip._get_address_lines` computes the address of the partner.
    
    This is done in an onchange context whereas the changes where not isolated
    and they were propagated to the real partner
    via all the onchanges related to it.
    
    We spotted the issue on the mail compose wizard
    whereas the chain of onchanges was wiping all the default values.
    
    With this change we:
    
    * isolate the partner w/ a fake record
    * additionally we cache the result per each partner
      since the function is called several times

### DIFF
--- a/l10n_ch_payment_slip/models/payment_slip.py
+++ b/l10n_ch_payment_slip/models/payment_slip.py
@@ -391,7 +391,7 @@ class PaymentSlip(models.Model):
                                        self.font_absolute_path()))
 
     @api.model
-    def _get_samll_text_font(self):
+    def _get_small_text_font(self):
         """Register a :py:class:`reportlab.pdfbase.ttfonts.TTFont`
         for recept reference
         :return: a :py:class:`FontMeta` with font name and size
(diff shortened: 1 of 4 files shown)


simahawk:
need to fix tests

l10n_ch_payment_slip/models/payment_slip.py
```diff
@@ -443,18 +443,22 @@ def _get_scan_line_text_font(self, print_settings):
...
+        # clone the partner to not affect real partner
+        com_partner = self.env['res.partner'].browse(com_partner_id)
+        partner = self.env['res.partner'].new(com_partner.copy_data()[0])
         # use onchange to define our own temporary address format
         with self.env.do_in_onchange():
             # assign a fake country in case partner has no country set
-            com_partner.country_id = self.env['res.country'].new()
-            com_partner.country_id.address_format = bvr_address_format
-            address_lines = com_partner._display_address(
+            partner.country_id = self.env['res.country'].sudo().new({
```
yvaucher: Do we really need the sudo ? does `new()` checks access rights?

yvaucher (approved):
Needs some tests otherwise LGTM.

l10n_ch_payment_slip/models/payment_slip.py
simahawk: @yvaucher was in the old code. Was remove recently indeed. Anyway it does not hurt here and I prefer to stay safe and not break the whole thing on customers' face :smile:

simahawk:
@yvaucher tests for this module are passing. Others are failing for apparently unrelated reasons. Can you check?

yvaucher:
Tests errors are indeed unrelated

PR #406, opened 2018-04-03
[IMP] l10n_ch_base_bank: bic and city required on view for switzerland

asaunier:
Replace PR #400 by @mpanarin 
See PR #405 for version 9.0 and #407 for version 11.0


asaunier:
@yvaucher Since you have merged PR #405 (which does basically the same for v9), do you approve the current PR?
Should I squash my fixup commit with @mpanarin 's?

coverage/coveralls seems to be stuck, but the other coverage checks seem OK.

yvaucher:
I'll squash them while merging

PR #408, opened 2018-04-04
[10.0][FIX] Fix pain_credit_transfer with multiple currencies

p-tombez:
Override base add-on to comply with Swiss Payment Standards : it is not allowed to have payments with different currencies in the same B-Level block.

Needs:
- [x] https://github.com/OCA/bank-payment/pull/462 to work

### DIFF
--- a/l10n_ch_pain_credit_transfer/models/account_payment_order.py
+++ b/l10n_ch_pain_credit_transfer/models/account_payment_order.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# copyright 2018 Camptocamp
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+from odoo import models, api
+
+
+class AccountPaymentOrder(models.Model):
+    _inherit = 'account.payment.order'
+
+    @api.model
+    def _get_line_key(self, line):
+        return super(AccountPaymentOrder, self)._get_line_key(line) + (
+            line.currency_id.name,
+        )
+
+    @api.model
+    def generate_start_payment_info_block(
+            self, parent_node, payment_info_ident,
+            priority, local_instrument, category_purpose, sequence_type,
+            requested_date, eval_ctx, gen_args):
+        return super(AccountPaymentOrder,
+                     self).generate_start_payment_info_block(
+            parent_node, "self.name + '-' "
+                         "+ requested_date.replace('-', '') + '-' "
+                         "+ priority + '-' "
+                         "+ optional_args[0] + '-' "
+                         "+ local_instrument + '-' "
+                         "+ category_purpose",
+            priority, local_instrument, category_purpose, sequence_type,
+            requested_date, eval_ctx, gen_args)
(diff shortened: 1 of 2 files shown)


l10n_ch_pain_credit_transfer/models/account_payment_order.py
```diff
@@ -0,0 +1,31 @@
...
+        )
+
+    @api.model
+    def generate_start_payment_info_block(
+            self, parent_node, payment_info_ident,
+            priority, local_instrument, category_purpose, sequence_type,
+            requested_date, eval_ctx, gen_args):
+        return super(AccountPaymentOrder,
+                     self).generate_start_payment_info_block(
+            parent_node, "self.name + '-' "
```
ghost: maybe better to computer 
```
 payment_info_ident = "self.name + '-' "
                         "+ requested_date.replace('-', '') + '-' "
                         "+ priority + '-' "
                         "+ optional_args[0] + '-' "
                         "+ local_instrument + '-' "
                         "+ category_purpose"
```
outside call, to increase readability?

leemannd (approved):
Not totally sure

leemannd:
@yvaucher What's your opinion on this?

l10n_ch_pain_credit_transfer/models/account_payment_order.py
yvaucher: I agree it would be cleaner to use a variable to avoid this strange indent as @oleksandrpaziuk mentionned.

@oleksandrpaziuk that code is just missing the wrapping '()'

l10n_ch_pain_credit_transfer/models/account_payment_order.py
```diff
@@ -0,0 +1,31 @@
...
+    def _get_line_key(self, line):
+        return super(AccountPaymentOrder, self)._get_line_key(line) + (
+            line.currency_id.name,
+        )
+
+    @api.model
+    def generate_start_payment_info_block(
+            self, parent_node, payment_info_ident,
+            priority, local_instrument, category_purpose, sequence_type,
+            requested_date, eval_ctx, gen_args):
```
yvaucher, line 21: Can you add a docstring to explain the change here where you prepare a more complex value for start_payment in order to differenciate from other blocks.

Also, I'm not sure of which part of the payment_info_ident attribute is going to be the currency itself.

yvaucher (changes requested):
A tiny refactor and a docstring and we are good here.

Thanks for the fix.

l10n_ch_pain_credit_transfer/models/account_payment_order.py
ghost: @yvaucher sorry I use code only for example purpose )
next time put the valid code into a comment )

yvaucher (approved):

l10n_ch_pain_credit_transfer/models/account_payment_order.py
p-tombez: Done

ecino:
Hello @p-tombez @yvaucher . Since the merge of this PR, we have crashes on our database when generating pain files. It seems that when evaluating the `payment_info_ident` variable, both values `optional_args` and `category_purpose` are not present in the `eval_context` and this is crashing the operation. Can I ask why is this inheritance overriding completely the `payment_info_ident`? I have the feeling that something is wrong with this way of doing and does not preserve well the behaviour across the inheritances of the function.

ecino:
I should add that we are generating postfinance dd files using the l10n_ch_lsv_dd module that calls directly the method `generate_start_payment_info_block` https://github.com/OCA/l10n-switzerland/blob/d3b94054863c9110c971e6e59880a64bcb6eb8e1/l10n_ch_lsv_dd/models/account_payment_order.py#L479

PR #409, opened 2018-04-10
[10.0][MIG] Migrate FDS modules from version 9.0

ecino:
This takes all FDS modules and push back the version 9.0, then add a migration commit to version 10.0

It adds as well a new module, l10n_ch_import_pain000 which is useful for the upload_sepa module. It allows parsing the return of Postfinance SEPA payment and automatically cancel payment lines that were rejected.


ecino:
@yvaucher All the long list of commits are commits that were made on the branch 9.0. Should we squash all these into one? The only useful commits to review here are the last two commits.

yvaucher:
@ecino you can squash commits by logic chunk. For instance all OCA Transbot can be in a single commit.

The ones about pep8 can be merged at least together by module.

If some are fixup (for instance typo or pep8) for an other commit you can squash them too.

ecino:
@yvaucher you can review bank-payment/433 if you have time, that would help for the tests to pass.

ecino:
@yvaucher The PR for FDS modules is finally good to go! Now you can review when you have time. Thanks

ecino:
Needs review

yvaucher (approved):

yvaucher:
Thanks for your work.

AFAIK we don't have any other module for pain 002 in the OCA. This one seems to be the swiss flavor, it might be interesting for other countries though. For now, let's keep it in this localization, I don't know what it would require to make it usable for the EU.

PR #411, opened 2018-04-11
[10.0] Allow payment slip to be saved in and read from attachment

grindtildeath:


ghost (approved):

l10n_ch_payment_slip/views/account_invoice.xml
```diff
@@ -1,71 +1,69 @@
...
-            <field name="reference"/>
-            <field name="move_line_id"/>
-            <field name="amount_total"/>
-            <field name="scan_line"/>
-            <field name="slip_image"/>
-            <field name="a4_pdf"/>
-          </group>
-        </form>
-      </field>
+        <field name="name">pay slip form</field>
```
simahawk: were these changes needed?

l10n_ch_payment_slip/report/report.py
```diff
@@ -17,40 +25,115 @@ class Report(models.Model):
...
+    def _generate_one_slip_per_page_from_invoice_pdf(self, report_name=None,
+                                                     save_in_attachment=None):
         """Generate payment slip PDF(s) from report model.
         If there is many pdf they are merged in memory or on
         file system based on company settings
 
         :return: the generated PDF content
         """
+        if not save_in_attachment:
+            save_in_attachment = {}
```
simahawk, line 37: state in a comment what you map here

l10n_ch_payment_slip/report/report.py
```diff
@@ -17,40 +25,115 @@ class Report(models.Model):
...
+
+        pdfdocuments = []
+        temporary_files = []
+
+        for doc in docs:
+            pdfreport_fd, pdfreport_path = tempfile.mkstemp(
+                suffix='.pdf', prefix='report.tmp.')
+            temporary_files.append(pdfreport_path)
+
+            if (
```
simahawk: split this method into pieces as it's hard to read, debug and maintain. You could split by concern, like `_handle_save_attachment` or alike.

l10n_ch_payment_slip/report/report.py
```diff
@@ -17,40 +25,115 @@ class Report(models.Model):
...
+            with closing(os.fdopen(pdfreport_fd, 'w')) as pdfreport:
+                pdf = doc._draw_payment_slip(a4=True, b64=False,
+                                             out_format='PDF',
+                                             report_name=report_name)
+                pdfreport.write(pdf)
+
+            if save_in_attachment and save_in_attachment.get(
+                    doc.invoice_id.id):
+                with open(pdfreport_path, 'rb') as pdfreport:
+                    attachment = {
```
simahawk: for instance, this could go into another smaller method `_create_attachment`

l10n_ch_payment_slip/report/report.py
```diff
@@ -17,40 +25,115 @@ class Report(models.Model):
...
+
+        return content
 
     @api.multi
     def get_pdf(self, docids, report_name, html=None, data=None):
         if (report_name == 'l10n_ch_payment_slip.'
                            'one_slip_per_page_from_invoice'):
+            report_model = self.env['ir.actions.report.xml']
+            context = self.env['res.users'].context_get()
+            report = report_model.with_context(context).search(
```
simahawk: do you want to replace the context completely? otherwise use `**context`

l10n_ch_payment_slip/report/report.py
```diff
@@ -60,11 +143,8 @@ def merge_pdf_in_memory(self, docs):
         streams = []
         writer = pyPdf.PdfFileWriter()
         for doc in docs:
-            current_buff = StringIO.StringIO()
-            streams.append(current_buff)
-            current_buff.write(doc)
-            current_buff.seek(0)
-            reader = pyPdf.PdfFileReader(current_buff)
+            pdfreport = file(doc, 'rb')
```
simahawk, line 155: so, `doc` here is a file content? if it's a path rename it to `filepath` or similar and use `open` and in a ctx manager or close it later

simahawk (changes requested):

l10n_ch_payment_slip/views/account_invoice.xml
grindtildeath: Not required, but since we had `openerp` and `data` element in v10, it make sense replacing with odoo. As it's one element instead of two, some reformatting make sense IMO.

grindtildeath:
@simahawk Most of the changes here are taken from odoo methods with some small adaptations, cf :
https://github.com/odoo/odoo/blob/10.0/addons/report/models/report.py#L373
https://github.com/odoo/odoo/blob/10.0/addons/report/models/report.py#L414
https://github.com/odoo/odoo/blob/10.0/addons/report/models/report.py#L456
https://github.com/odoo/odoo/blob/10.0/addons/report/models/report.py#L561

I'm not sure it makes much sense spending time to split this process, as this concerns only the payment slip feature, which is already included in odoo 11, so it won't have to be ported...

l10n_ch_payment_slip/views/account_invoice.xml
yvaucher: Please don't change from 2 spaces to 4. 2 spaces is a valid in OCA and makes it more readable on large views and saves few chars. (there has been a discution on the mailing lists long time ago without a decision taken)

It would also make it harder to apply fixes on multiples versions.

yvaucher (changes requested):
Don't change indent in xml files

l10n_ch_payment_slip/views/account_invoice.xml
grindtildeath: Then, what are we supposed to do when changing  `<openerp> <data>` to `<odoo>` ? 
Leave 4 spaces after first element (odoo) and 2 afterwards, or reindent with 2 spaces ?

yvaucher: Indeed, you are right it would break for cherry-pick anyway.

Still, please keep the 2 spaces. It is still readable and save space in view definition that can have several level of xml. And when opening a file and seeing only spaces it doesn't help.

Or please explain why you want to change it.

grindtildeath:
@yvaucher Fixed and rebased.

grindtildeath:
@yvaucher @simahawk Can you guys please update your reviews ? :)

yvaucher (approved):

simahawk (approved):

PR #413, opened 2018-04-13
[10.0] l10n_ch_scan_bvr : Consider bank account of parent partner

grindtildeath:


simahawk (approved):

fclementic2c:
Why dont you use the partner commercial_id method.
if you have a hierarchy of 3 levels you're fu*** don't you think ?

grindtildeath:
@fclementic2c It seems we don't need to use the `commercial_partner_id` in this case, as the `parent_of` operator will search through multiple parent_id relations. What these modifications do, are :
1. Allow to select the bank account of any parent partner on the wizard's form.
2. Create the invoice with the partner chosen on the wizard instead of the partner (parent) linked to the bank account selected.
3. Select the default product from the partner of the invoice (in case it's the same as bank account) or fall back on the default product from the partner of the bank account (which is the parent and the commercial_partner_id anyway).

fclementic2c:
:+1: 
Ok for me

PR #417, opened 2018-04-22
[11.0][FIX] _compute_acc_type failing on recordsets

mpanarin:
As the method is multi, it shouldn't use self to get record fields.
@yvaucher


yvaucher:
This one is trivial -> merged

Issue #396, opened 2018-02-19
[9.0] l10n_ch_fds_postfinance cannot properly connect to bank


bodi000:
Hello,

I installed the module and configured an existing fds account. I can connect to it via ssh in ubuntu. The keys are valid. I added the public key to the known_hosts file. 
1. While trying to connect via the Import from FDS wizard, an error is thrown: SFTP unable to connect.

in the log I get the following message: 
``2018-02-19 10:40:02,686 1627 ERROR bs_06_v8_migrated openerp.addons.l10n_ch_fds_postfinance.wizards.fds_files_import_tobankstatments_wizard: None``

2. For the Verify folders wizard, the warning " Unable to connect to the sftp " shows-up.
the log has:
``2018-02-19 10:41:17,450 1627 ERROR bs_06_v8_migrated openerp.addons.l10n_ch_fds_postfinance.models.fds_postfinance_account: Unable to connect to the sftp: No hostkey for host fdsbc.post.ch found.``

Surprisingly, while running through eclipse debug (same config file), I was able to connect and download the files from FDS, so I could not find the error.

Any help is welcome!
Thank you
b

bodi000:
It appears that the problem comes from ssh keys [not] available to odoo user which runs odoo.

yvaucher:
@ecino any idea on this?

ecino:
Hi @bodi000, please verify that you configured correctly your users in the Accounting->Configuration->FDS Postfinance Acccount menu.
You should add the ssh public key your are using for each user that will connect to FDS. See the attached screenshot:
![fds-config](https://user-images.githubusercontent.com/8435180/39117658-c63ac530-46e7-11e8-87be-9c51355c56a6.jpg)

bodi000:
Hello and thank you @ecino for your answer. 
As I mentioned, everything works fine in the pydev debugger (local user) but not directly (dedicated user). It also works (i can connect to fds) in external applications, e.g. fileZilla, so I think that there is some issue with ssh configuration for the openerp dedicated system user. I tried to configure /etc/ssh/ssh_config to use a specific private key, maybe something is missing...

Added:
```
Host fdsbc.post.ch
     HostName fdsbc.post.ch
     HashKnownHosts yes
     IdentityFile /etc/ssh/asio_id_rsa
```
before Host *

Thank you in advance

davidwul:
Hi @bodi000 I think I had this issue. One way to solve that is to connect to fds:
[...]

PR #418, opened 2018-04-24
[IMP] workaround for finding known_hosts file (#396)

bodi000:


yvaucher:
@bodi000 can you detail a bit more what your intent is here?
Do you have SSL issues?

l10n_ch_fds_postfinance/models/fds_postfinance_account.py
```diff
@@ -103,10 +103,13 @@ def verify_directories_button(self):
                 key[0].private_key_crypted)
             key_pass = self.authentication_key_ids.config()
 
+            cnopts = pysftp.CnOpts(knownhosts='/etc/ssh/ssh_known_hosts')
```
ecino: I fear this will fail on a non-linux host.

ecino (commented):
In the documentation of pysftp, changing the knownhosts is not recommended:
image
Are you sure there is no other fix for this? Normally you should connect with ssh outside odoo, with the odoo user to the server at least one time to add the host to your known hosts.

bodi000:
Hello. Yes, as it is explained in the issue #396...
However, the behaviour comes from a bug at pysftp level. I filed an issue there and a PR(which I need to rework ;))

This is just the workaround for using a system user without HOME folder.

If the upstream bug gets solved, no need for this PR.

Thank you for your understanding
B

l10n_ch_fds_postfinance/models/fds_postfinance_account.py
bodi000: Hello. As I mentioned I use a system user. No Home folder, so no usefullness into connecting once.

Please park this for a few days, i am tryng to solve this with pysftp directly. My issue was recongnised as a bug there ..

PR #425, opened 2018-05-23
[11.0][FIX] l10n_ch_pain_base - fix error message

lfr-opennet:
Fix of the field used on the UsserError message when no CCP account is set.
The field 'partner_bank.name' doesn't exist, the corresponding field is 'partner_bank.bank_id.name'

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -89,7 +89,7 @@ class AccountPaymentOrder(models.Model):
             if not partner_bank.ccp:
                 raise UserError(_(
                     "The field 'CCP/CP-Konto' is not set on the bank "
-                    "account '%s'.") % partner_bank.name)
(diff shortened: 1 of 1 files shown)


l10n_ch_pain_base/models/account_payment_order.py
```diff
@@ -89,7 +89,7 @@ def generate_party_acc_number(self, parent_node, party_type, order,
             if not partner_bank.ccp:
                 raise UserError(_(
                     "The field 'CCP/CP-Konto' is not set on the bank "
-                    "account '%s'.") % partner_bank.name)
+                    "account '%s'.") % partner_bank.bank_id.name)
```
yvaucher: I think it should be `partner_bank.acc_number` instead. bank_id could even not be set.
The message is there to help identify which bank account is messing up with the pain generation.

yvaucher (changes requested):

yvaucher (approved):

PR #426, opened 2018-05-25
[FIX] make only 1 payment

vrenaville:
This module previously create one payment operation for each line
it generate a lot of document.

I have also a add the tag InstrID, it's optional for SIX payment validation tools, but required by most cantonal banks


leemannd (approved):
Seems legit.

yvaucher (approved):
Indeed much better, thanks for the improvement

PR #428, opened 2018-06-13
[11.0][FIX] res.partner.bank: retrieve partner name from context on creation

MaxyMoos:
This commit fixes a minor issue where creating a `res.partner.bank` instance from a partner form (_Accounting_ > _Master Data_ > _Customers/Vendors_ > _Sales & Purchases_ > _Bank accounts_  menu entries) and entering a CCP code (e.g. _46-110-7_) would automatically generate an account name: "_Bank/CCP Undefined_" when we would expect the related partner name to be present.

As the related partner ID is available in the context, we retrieve it from there.

### DIFF
--- a/l10n_ch_base_bank/models/bank.py
+++ b/l10n_ch_base_bank/models/bank.py
@@ -239,11 +239,16 @@ class ResPartnerBank(models.Model, BankCommon):
     def _get_acc_name(self):
         """ Return an account name for a bank account
         to use with a ccp for ISR.
-        This method make sure to generate a unique name
+        This method makes sure to generate a unique name
         """
+
         part_name = self.partner_id.name
+        if not part_name and self.env.context.get('default_partner_id'):
+            partner_id = self.env.context.get('default_partner_id')
+            part_name = self.env['res.partner'].browse(partner_id)[0].name
+
         if part_name:
-            acc_name = _("Bank/CCP {}").format(self.partner_id.name)
+            acc_name = _("Bank/CCP {}").format(part_name)
         else:
             acc_name = _("Bank/CCP Undefined")
 
--- a/l10n_ch_base_bank/__manifest__.py
+++ b/l10n_ch_base_bank/__manifest__.py
@@ -3,7 +3,7 @@
 
 {'name': 'Switzerland - Bank type',
  'summary': 'Types and number validation for swiss electronic pmnt. DTA, ESR',
- 'version': '11.0.1.1.0',
+ 'version': '11.0.1.1.1',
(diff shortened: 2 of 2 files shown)


l10n_ch_base_bank/models/bank.py
```diff
@@ -239,11 +239,16 @@ def _check_postal_num(self):
     def _get_acc_name(self):
         """ Return an account name for a bank account
         to use with a ccp for ISR.
-        This method make sure to generate a unique name
+        This method makes sure to generate a unique name
         """
+
         part_name = self.partner_id.name
+        if not part_name and self.env.context.get('default_partner_id'):
+            partner_id = self.env.context.get('default_partner_id')
```
yvaucher, line 247: I don't see the `default_partner_id` being set in the `res.partner.bank` view.

While I like the idea, is it a customization on your side? How is odoo defining that `default_partner_id` ?

yvaucher:
In version 11.0 I see the following path for Bank accounts:

Contacts / Configuration / Bank Accounts / Bank Accounts
Accounting / Configuration / Accounting / Bank Accounts 

Which doesn't seems to match yours.

l10n_ch_base_bank/models/bank.py
MaxyMoos: Hi @yvaucher, thanks for the reply.

The path I use to create/access partner bank accounts is located in a partner form view (_Invoicing_ > _Sales_ > _Master Data_ > _Customers_ > open a customer > tab _Sales & Purchases_ > _Bank Account(s)_)

`default_partner_id` is set in `odoo/addons/account/views/partner_views.xml`, around line 124 from what I can see :

```
                <group name="container_row_2" position="inside">
                    <group string="Payments" name="banks">
                        <field name="currency_id" invisible="1"/>
                        <div>
                            <button type="action" class="btn-link"
                                name="%(base.action_res_partner_bank_account_form)d"
                                context="{'search_default_partner_id': active_id, 'default_partner_id': active_id}">
                                <field string="Bank account(s)" name="bank_account_count" widget="statinfo"/>
                            </button>
                        </div>

                    </group>
                </group>
```

MaxyMoos:
@yvaucher I resolved the opened conversation with details on the `default_partner_id` field but don't think I can remove the 'Question' tag on the issue to highlight resolution.
If you need more details, I'll be glad to help !

yvaucher:
Ok misread your description, you meant, in 

> Partner Form > Tab Sales & Purchases > Button Bank accounts

yvaucher (approved):
This works as expected tested locally, thanks!

PR #431, opened 2018-07-04
FIX postfinance attachment parser

ecino:
I made a previous correction on the attachments parser in 9708dec1cb95835fa5f6dbfcf626c7727737f405.
While this was an improvement, it is not totally correct as some transactions don't have the same instrid and file name. It results as some images not being attached to the statement.

With this fix, all the images are now correctly attached to the statement.


yvaucher:
@ecino are you adding the file l10n_ch_bank_statement_import_postfinance/test_files/demo_pf_ch.tar.gz ?

Not sure why it is displayed in the commit diff.

ecino:
I changed the test file because it was not containing the xml tag in which is found the Prtry Ref.

yvaucher (approved):

yvaucher:
I merge and will bump the version number

PR #434, opened 2018-07-24
wrong field name

Martronic-SA:
It seems that the field "name" doesn't exists in res.partner.bank model. It's "bank_name" instead.

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -90,7 +90,7 @@ class AccountPaymentOrder(models.Model):
             if not partner_bank.ccp:
                 raise UserError(_(
                     "The field 'CCP/CP-Konto' is not set on the bank "
-                    "account '%s'.") % partner_bank.name)
+                    "account '%s'.") % partner_bank.bank_name)
             party_account = etree.SubElement(
                 parent_node, '%sAcct' % party_type)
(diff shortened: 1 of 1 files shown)


l10n_ch_pain_base/models/account_payment_order.py
```diff
@@ -90,7 +90,7 @@ def generate_party_acc_number(
             if not partner_bank.ccp:
                 raise UserError(_(
                     "The field 'CCP/CP-Konto' is not set on the bank "
-                    "account '%s'.") % partner_bank.name)
+                    "account '%s'.") % partner_bank.bank_name)
```
yvaucher, line 93: You don't want bank_name here, you want acc_number to be relevant. If you had multiple account on the same bank you wouldn't know where the error message came from.

This should have been addressed already in https://github.com/OCA/l10n-switzerland/pull/425

I guess the error in multiple odoo versions.

I'll backport the fix.

yvaucher (changes requested):

yvaucher:
fixed here: https://github.com/OCA/l10n-switzerland/commit/3902076e5836eef94be3997ab43f416d734377ac

yvaucher:
Thanks for reporting it, I also checked version 10.0 and this was fixed a long time ago:
https://github.com/OCA/l10n-switzerland/pull/383

PR #373, opened 2017-11-17
[MIG] l10n_ch_payment_slip: Migration to 11.0

yvaucher:
Update headers
Port code to python3
Update README
Fix tests with correct report render calls
Adapt code to report module split

TODO: 
- [x] fixing tests
- [x] v11 + tests -> moved in another module
- [x] put bvr parameters on res.confing.settings
- [ ] edit README.rst to explain the difference with Odoo official addon.

PR #435, opened 2018-07-24
[11.0][IMP] Remove useless button on invoice

p-tombez:
Remove the button "Print ISR" on invoices to avoid confusion with Odoo base
implementation and rename the one from this module for user consistency

### DIFF
--- a/l10n_ch_payment_slip/views/account_invoice.xml
+++ b/l10n_ch_payment_slip/views/account_invoice.xml
@@ -1,74 +1,74 @@
 <?xml version="1.0"?>
 <openerp>
-    <data>
+  <data>
 
-        <record id="action_slip_all" model="ir.actions.act_window">
-            <field name="name">Payment Slip</field>
-            <field name="res_model">l10n_ch.payment_slip</field>
-            <field name="domain">[('invoice_id', '=', active_id)]</field>
-        </record>
+    <record id="action_slip_all" model="ir.actions.act_window">
+      <field name="name">Payment Slip</field>
+      <field name="res_model">l10n_ch.payment_slip</field>
+      <field name="domain">[('invoice_id', '=', active_id)]</field>
+    </record>
 
-        <record model="ir.ui.view" id="invoice_form_add_bvr_ref">
-            <field name="name">account.invoice.form.bvr.ref</field>
-            <field name="model">account.invoice</field>
-            <field name="type">form</field>
-            <field name="inherit_id" ref="account.invoice_form"/>
-            <field name="arch" type="xml">
-                <xpath expr="//button[@name='invoice_print'][1]" position="after">
-                    <button type="object"
-                            string="Print CH ISR"
-                            name="print_bvr"
-                            attrs="{'invisible': ['|',('sent','=',True), ('state', '!=', 'open')]}"
-                            class="oe_highlight"/>
-                </xpath>
-                <xpath expr="//button[@name='invoice_print'][2]" position="after">
-                    <button type="object"
-                            string="Print CH ISR"
-                            name="print_bvr"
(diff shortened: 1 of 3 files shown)


ghost (approved):
LGTM.
Try to separate changes and indentation fix in two commits.

asaunier:
With the indentation changes, it is not easy to spot the other changes.

p-tombez:
@oleksandrpaziuk @asaunier You can hide the indentation diff when you review the changes :
image

asaunier:
@p-tombez That's much clearer, thanks :)
The changes looks OK to me but I have no clue if it is OK to remove the button.

leemannd:
@p-tombez Is the `isr_print` always present? If not, won't it fail?

@asaunier It is recent functionnality ;)

l10n_ch_payment_slip/views/account_invoice.xml
```diff
@@ -1,71 +1,72 @@
...
+                            attrs="{'invisible': ['|',('sent','=',True), ('state', '!=', 'open')]}"
+                            class="oe_highlight"/>
+                </xpath>
+                <xpath expr="//button[@name='invoice_print'][2]" position="after">
+                    <button type="object"
+                            string="Print CH ISR"
+                            name="print_bvr"
+                            attrs="{'invisible':['|',('sent','=',False), ('state', '!=', 'open')]}"/>
+                </xpath>
+                <xpath expr="//button[@name='isr_print']" position="replace"/>
```
Tonow-c2c: Why you prefer to remove and not to put to invisible ?

p-tombez: Does it really make a difference ? The goal is to not be able to use that button if this module is installed

leemannd: @p-tombez If you have another module extending it the xpath expression will fail correct?

yvaucher:
@p-tombez can you add a note in README.md that this module superseed the core l10n_ch module for ISR?

l10n_ch_payment_slip/views/account_invoice.xml
Tonow-c2c: @leemannd Yes I think

p-tombez:
@leemannd Yes, it's always present through dependencies
@Tonow-c2c Made it invisible instead of removing it
@yvaucher Note added

leemannd (approved):
Thanks

l10n_ch_payment_slip/views/account_invoice.xml
```diff
@@ -1,71 +1,74 @@
...
 <openerp>
-  <data>
+    <data>
 
-    <record id="action_slip_all" model="ir.actions.act_window">
-      <field name="name">Payment Slip</field>
-      <field name="res_model">l10n_ch.payment_slip</field>
-      <field name="domain">[('invoice_id', '=', active_id)]</field>
-    </record>
+        <record id="action_slip_all" model="ir.actions.act_window">
```
yvaucher: Please don't change the indentation from 2 spaces to 4 spaces. That makes xml big for nothing and history is overrided.

yvaucher (changes requested):

l10n_ch_payment_slip/views/account_invoice.xml
p-tombez: The indentation in this file was an absolute mess. I can change to have 2 spaces instead of 4 to avoid big files but there will still be indentation changes.

yvaucher (approved):

yvaucher:
@p-tombez could you propose this on top of https://github.com/OCA/l10n-switzerland/pull/373 ?

PR #436, opened 2018-07-26
[ADD] line.communication of paiement even if it's not a bvr reference

vrenaville:


nbessi (approved):
LGTM seems legit

CharlineDumontet (approved):

l10n_ch_sepa/l10n_ch/template/pain.001.001.03.ch.02.xml.mako
```diff
@@ -58,15 +58,13 @@
...
-          <RmtInf>
-            <Strd>
-              <CdtrRefInf>
-                <Ref>${line.communication.replace(' ', '')}</Ref>
-              </CdtrRefInf>
-            </Strd>
-          </RmtInf>
-   % endif
+  <RmtInf>
+    <Strd>
```
yvaucher, line 62: Strd is for structured references only. I don't think we should pass a free reference that way.

yvaucher (changes requested):

l10n_ch_sepa/l10n_ch/template/pain.001.001.03.ch.02.xml.mako
yvaucher: Ok it must be BVR only for type 1 which is what was implemented.

For other types, it is a structured reference like ISO 11649, QR or IPI. 

Page 44
https://www.six-interbank-clearing.com/dam/downloads/fr/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf

```
2.126 Ref:

Type 1: Doit contenir le numéro de référence BVR.
Type 5: Doit être 
utilisée si «Creditor Reference
Information» est utilisé, peux contenir «Creditor
Reference» conformément à ISO 11649.
Autres types: Références structurées, par exemple
référence ISO (ISO 11649) ou référence QR ou
référence IPI.
Remarque: Le justificatif I
PI doit être retiré de la
circulation au plus tard le 31.03.2020.
```

More information about reference on page 46

vrenaville:
@yvaucher 

You are right, I will check to add a free reference.

Vincent

PR #446, opened 2018-09-27
[8.0] [IMP] IBAN is mandatory for account if we set an other account it wil…

vrenaville:
…l not work


yvaucher:
@vrenaville you have a conflict in mako file

PR #447, opened 2018-10-03
[10.0] l10n_ch_scan_bvr - Support of Postal BVR

Martronic-SA:
It's possible to have BVR for Postal accounts that are type 01 but with 53 positions.
Now your are getting the bvrnumber (bvr adherent number) and searching for it but they have no bvr adherent number, only the CCP.
By the way the bvrnumber for those BVR is not equal to '000000'.
Then if there is something in the bvrnumber, you should search for CCP/adherent = bvrnumber or CCP only (adherent = False)

### DIFF
--- a/l10n_ch_scan_bvr/wizard/scan_bvr.py
+++ b/l10n_ch_scan_bvr/wizard/scan_bvr.py
@@ -353,8 +353,9 @@ class ScanBvr(models.TransientModel):
             domain = [('ccp', '=', data['bvr_struct']['beneficiaire'])]
         else:
             domain = \
-                [('ccp', '=', data['bvr_struct']['beneficiaire']),
-                 ('bvr_adherent_num', '=', data['bvr_struct']['bvrnumber'])]
+                [('ccp', '=', data['bvr_struct']['beneficiaire']), '|',
+                 ('bvr_adherent_num', '=', data['bvr_struct']['bvrnumber']),
+                 ('bvr_adherent_num', '=', False)]
         partner_bank = partner_bank_model.search(domain, limit=1)
         # We will need to know if we need to create invoice line
         if partner_bank:
--- a/l10n_ch_scan_bvr/wizard/scan_bvr.py
+++ b/l10n_ch_scan_bvr/wizard/scan_bvr.py
@@ -349,14 +349,19 @@ class ScanBvr(models.TransientModel):
         partner_bank_model = self.env['res.partner.bank']
         partner_bank = False
         # We will now search the account linked with this BVR
-        if data['bvr_struct']['domain'] == 'name':
-            domain = [('ccp', '=', data['bvr_struct']['beneficiaire'])]
-        else:
-            domain = \
-                [('ccp', '=', data['bvr_struct']['beneficiaire']), '|',
-                 ('bvr_adherent_num', '=', data['bvr_struct']['bvrnumber']),
-                 ('bvr_adherent_num', '=', False)]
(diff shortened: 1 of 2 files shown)


l10n_ch_scan_bvr/wizard/scan_bvr.py
```diff
@@ -353,8 +353,9 @@ def validate_bvr_string(self):
             domain = [('ccp', '=', data['bvr_struct']['beneficiaire'])]
         else:
             domain = \
-                [('ccp', '=', data['bvr_struct']['beneficiaire']),
-                 ('bvr_adherent_num', '=', data['bvr_struct']['bvrnumber'])]
+                [('ccp', '=', data['bvr_struct']['beneficiaire']), '|',
+                 ('bvr_adherent_num', '=', data['bvr_struct']['bvrnumber']),
+                 ('bvr_adherent_num', '=', False)]
```
yvaucher: @Martronic-SA Shouldn't we refactor it to first search for a matching `bvr_adherent_num` then if not found search for a `bvr_adherent_num` set to false?

Plus, what if the ccp field is present on multiple accounts in your ERP? (with Bank's CCP)

You could end up matching the wrong supplier.

yvaucher (changes requested):

l10n_ch_scan_bvr/wizard/scan_bvr.py
Martronic-SA: This has been resolved by the commit https://github.com/OCA/l10n-switzerland/pull/447/commits/a8a8d316f5730525897a20a414169e23b738b6d9 below.

yvaucher:
> It's possible to have BVR for Postal accounts that are type 01 but with 53 positions.

What do you mean by this? ISR ref are max 27 positions, and postal account number are 9 digits length (there was an old format with 6 but is obsolete).
Is `53` a typo?

Other than that I understand the module currently only works for ISR-B and not ISR for postal account.
And this patch should do the trick.

And side note this patch seems important as even if QR-Invoice will arrive soon, we can still receive ISR payment slips till 2022.

Martronic-SA:
> What do you mean by this? ISR ref are max 27 positions, and postal account number are 9 digits length (there was an old format with 6 but is obsolete).
> Is `53` a typo?

No in fact I wasn't speaking about the ISR ref itself, but the whole scanline thant can be 53 chars.

yvaucher:
@Martronic-SA Can you add a unit test with an example in
https://github.com/OCA/l10n-switzerland/blob/10.0/l10n_ch_scan_bvr/tests/test_scan_bvr.py ?

You can copy the test https://github.com/OCA/l10n-switzerland/blob/10.0/l10n_ch_scan_bvr/tests/test_scan_bvr.py#L73 and change the bvr_string by your case.

yvaucher:
set as `needs fixing` for additional unit test

Martronic-SA:
> 
> 
> set as `needs fixing` for additional unit test

I've updated the testings. I don't know where to check the coverage here but that should be ok.

yvaucher:
Indeed the travis configuration is missing coverage since a while. 

Last coverage was in nov 2019. I need to fix this.
https://coveralls.io/github/OCA/l10n-switzerland?branch=10.0

yvaucher (approved):
@Martronic-SA Thanks for the unit test.

Can you just squash your last commit please to keep the history cleaner?

Martronic-SA:
@yvaucher Is that ok now?

PR #449, opened 2018-10-12
[9.0][IMP] l10n_ch_bank_statement_import_postfinance add date from PF into the bank statement

bodi000:
The date from the PF statement is lost, as the file format diverges from the default one. This commit parses the correct date into account.bank.statement table.


bodi000:
Apparently it doesn't work...

bodi000:
I closed it as too many changes were still needed for travis. the new one is #451

PR #450, opened 2018-10-15
Error when using BVR in attachment

Martronic-SA:
When you are on an invoice, you might have and default value for the context of the action with {'default_type':'out_invoice'}.
This default value will be passed to the context of the ir.attachment and there is a field "type" in the attachment which allows "url" and "binary" as values. Then it will throw an error
>UserError: (Wrong value for ir.attachment.type: 'out_invoice'
We can force the type to be 'binary'

### DIFF
--- a/l10n_ch_payment_slip/report/report.py
+++ b/l10n_ch_payment_slip/report/report.py
@@ -81,6 +81,7 @@ class Report(models.Model):
                             doc.invoice_id.id),
                         'res_model': save_in_attachment.get('model'),
                         'res_id': doc.invoice_id.id,
+                        'type':'binary'
                     }
                     try:
                         self.env['ir.attachment'].create(attachment)


yvaucher (approved):
Interesting side effect of the context forwarding.

By default the type is `binary` so this patch just makes it explicit without a use of context.

yvaucher:
The code has been refactored, thus I cannot forward port this patch as is into 11.0 and 12.0.
It will only land in 10.0

l10n_ch_payment_slip/report/report.py
```diff
@@ -81,6 +81,7 @@ def _generate_one_slip_per_page_from_invoice_pdf(self, report_name=None,
                             doc.invoice_id.id),
                         'res_model': save_in_attachment.get('model'),
                         'res_id': doc.invoice_id.id,
+                        'type':'binary'
```
yvaucher, line 84:
```suggestion
                        'type': 'binary'
```

Just a missing space in fact.

yvaucher:
manually merged at https://github.com/OCA/l10n-switzerland/commit/a5f0c210027509cc2b7f26ea3d33666a0df871ee

Martronic-SA:
Perfect, thank you!

PR #456, opened 2018-12-10
[11.0] Enable SEPA payments when using Swiss pain.001 version

MaxyMoos:
Using the l10n_ch_pain_credit_transfer module, it is currently impossible to make SEPA orders using the Swiss pain version `pain.001.001.03.ch.02 (credit transfer in Switzerland)` while they are supported, according to the [SIX implementation guidelines](https://www.six-group.com/interbank-clearing/dam/downloads/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf) (see p.18):

This PR aims at enabling SEPA ("Type 5") for payment orders that can support it, instead of "Type 6" payments, as would currently be generated.

### DIFF
--- a/l10n_ch_pain_base/README.rst
+++ b/l10n_ch_pain_base/README.rst
@@ -36,6 +36,7 @@ Contributors
 * Alexis de Lattre <alexis.delattre@akretion.com>
 * Denis Leemann <denis.leemann@camptocamp.com>
 * Mykhailo Panarin <m.panarin@mobilunity.com>
+* Maxence Groine <mgroine@fiefmanage.ch>
 
 Maintainer
 ----------
--- a/l10n_ch_pain_base/__manifest__.py
+++ b/l10n_ch_pain_base/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Switzerland - ISO 20022",
     "summary": "ISO 20022 base module for Switzerland",
-    "version": "11.0.1.0.0",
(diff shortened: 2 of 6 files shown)


BT-astauder:
@MaxyMoos Thank you a lot for this contribution. When I understand the effect correctly this would mean that a payment order is either SEPA or not. Especially in Switzerland this is not a normal procedure as SEPA and non-SEPA transactions are mixed usually. Am I correct with that?

MaxyMoos:
@BT-astauder Thanks for your message & feedback.

The fact that a payment order is either SEPA or not is already coded at the [account_banking_pain_base](https://github.com/OCA/bank-payment/blob/11.0/account_banking_pain_base/models/account_payment_order.py) level, and I tried to change the smallest amount of code possible to make it possible to support SEPA payments.

I do agree that this makes it necessary for users to filter & group the payments they want to make into both SEPA-compliant payment orders and non-SEPA-compliant POs, which is not ideal given that I don't see the SIX guidelines explicitly forbidding mixed (SEPA + non-SEPA) B-levels.

If rewriting all the OCA modules from the top down in order to support different payment types in a single file should be the goal, we can drop this PR - it is indeed more of a fix for Swiss users !

yvaucher (approved):
LGTM, this fixes the same issue mentionned in https://github.com/OCA/l10n-switzerland/pull/487

But this patch takes care of domestic payment with EUR and IBAN (`Type 3`)

PR #458, opened 2018-12-14
[8.0][Fix] PAIN ZKB: InstrID recommended, but mandatory fpr ZKB

stephanamann:
The Element InstrID is not created, but ZKB requires it.

(Please ignore the commit from Jan. 03, this one has already been merged)


stephanamann:
Sorry, I'm struggling with using pull requests the right way :-( - I'm closing pull request and will resend it properly.

PR #465, opened 2019-02-08
[12.0][MIG] l10n_ch_pain_credit_transfer

asaunier:
Migrate also dependences ``l10n_ch_pain_base`` and ``l10n_ch_base_bank`` (dependence  [account_banking_sepa_credit_transfer](https://github.com/OCA/bank-payment/tree/12.0/account_banking_sepa_credit_transfer) is already migrated).

PR #470, opened 2019-02-27
[12.0] [MIG] l10n_ch_payment_slip  + l10n_ch_base_bank

i-vyshnevska:
pick latest changes from v 11.0 and update to v12


i-vyshnevska:
fails due to  https://github.com/OCA/l10n-switzerland/pull/465
and https://github.com/odoo/odoo/pull/31293

l10n_ch_payment_slip/report/ir_action.py
```diff
@@ -1,46 +1,172 @@
...
+        if res_ids:
+            _logger.info(
+                'The PDF report has been generated for records %s.' % (
+                    str(res_ids))
             )
+            return self._post_pdf(save_in_attachment, pdf_content=pdf_content,
+                                  res_ids=res_ids), 'pdf'
+        return pdf_content, 'pdf'
+
+    # TODO consider https://github.com/OCA/reporting-engine/issues/241
```
Tonow-c2c: Time to do this TODO ? [https://github.com/OCA/reporting-engine/issues/241](https://github.com/OCA/reporting-engine/issues/241)
or remove if it's no more relevant

yvaucher: This payment slip is about to die, the new format using QR code will be used in few months.

I think we can drop the TODO.

l10n_ch_bank/__init__.py
```diff
@@ -1,12 +1,8 @@
 # Copyright 2015 Yannick Vaucher (Camptocamp SA)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 from . import models
+from .hooks import import_csv_data
 
 
 def post_init(cr, registry):
```
simahawk: `post_init` should stay in `hooks.py` and simply be imported here

l10n_ch_base_bank/__manifest__.py
```diff
@@ -1,21 +1,24 @@
...
+# Copyright 2012-2019 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 {'name': 'Switzerland - Bank type',
  'summary': 'Types and number validation for swiss electronic pmnt. DTA, ESR',
- 'version': '11.0.1.1.0',
+ 'version': '12.0.1.0.0',
  'author': "Camptocamp,Odoo Community Association (OCA)",
  'category': 'Localization',
  'website': 'http://www.camptocamp.com',
```
simahawk: https://github.com/OCA/l10n-switzerland

l10n_ch_base_bank/__manifest__.py
```diff
@@ -1,21 +1,24 @@
...
- 'depends': ['account_payment_partner', 'base_iban'],
+ 'depends': [
+     'account_payment_order',
+     'account_payment_partner',
+     'base_iban'],
  'data': [
      'views/bank.xml',
      'views/invoice.xml',
  ],
  'demo': [],
```
simahawk: remove empty/default keys

l10n_ch_base_bank/models/bank.py
```diff
@@ -152,8 +152,8 @@ def name_search(self, name, args=None, operator='ilike', limit=80):
             ids = self.search(
                 args,
                 limit=limit
-                ).ids
```
simahawk, line 155: @yvaucher can't we build a big domain search for all the cols in OR and do only one search?

l10n_ch_base_bank/models/bank.py
```diff
@@ -236,29 +272,53 @@ def _check_postal_num(self):
...
+        """Check if number generated from ccp, if yes replace it on new """
         part_name = self.partner_id.name
-        if part_name:
-            acc_name = _("Bank/CCP {}").format(self.partner_id.name)
+        if not part_name and self.env.context.get('default_partner_id'):
+            partner_id = self.env.context.get('default_partner_id')
+            part_name = self.env['res.partner'].browse(partner_id)[0].name
+        self.acc_number = self._compute_name_ccp(part_name, self.ccp)
+
+    @api.multi
```
simahawk: seems more `api.model`

l10n_ch_base_bank/models/bank.py
```diff
@@ -303,58 +365,65 @@ def onchange_acc_number_set_swiss_bank(self):
...
-            if not self.acc_number and self.ccp:
-                if self.bank_id.is_swiss_post():
-                    self.acc_number = self.ccp
-                else:
-                    self.acc_number = self._get_acc_name()
+    def onchange_ccp_set_acc_number(self):
+        """If ccp changes and it's a postal bank update acc_number to ccp
+        we don't want make acc_number as computed to have possibility set it
+        manually and also avoid to shadow other logic on acc_number if exist
+       """
```
simahawk: indentation seems weird here

l10n_ch_base_bank/models/bank.py
```diff
@@ -303,58 +365,65 @@ def onchange_acc_number_set_swiss_bank(self):
...
-    def onchange_ccp_set_empty_acc_number(self):
-        """ If acc_number is empty and bank ccp is defined fill it """
-        if self.bank_id:
-            if not self.acc_number and self.ccp:
-                if self.bank_id.is_swiss_post():
-                    self.acc_number = self.ccp
-                else:
-                    self.acc_number = self._get_acc_name()
+    def onchange_ccp_set_acc_number(self):
+        """If ccp changes and it's a postal bank update acc_number to ccp
```
simahawk: this is more a comment than a docstring

l10n_ch_base_bank/models/bank.py
```diff
@@ -303,58 +365,65 @@ def onchange_acc_number_set_swiss_bank(self):
...
-                                            limit=1))
-            if not self.acc_number:
-                if not bank.is_swiss_post():
-                    self.acc_number = self._get_acc_name()
-                else:
-                    self.acc_number = self.ccp
+        if ccp and self.is_swiss_postal_num(ccp) and not self.bank_id.id:
+            bank = (self.env['res.bank'].search([('ccp', '=', ccp)], limit=1)
+                    or
+                    self.env['res.bank'].search([('bic', '=', 'POFICHBEXXX')],
```
simahawk: this hardcoded value comes from?

l10n_ch_base_bank/models/bank.py
```diff
@@ -303,58 +365,65 @@ def onchange_acc_number_set_swiss_bank(self):
...
-            if not self.acc_number and self.ccp:
-                self.acc_number = self._get_acc_name()
-            elif self.acc_number and self.is_swiss_postal_num(self.acc_number):
-                self.ccp = self.acc_number
-                self.acc_number = self._get_acc_name()
+            self._update_acc_name()
 
     @api.onchange('partner_id')
     def onchange_partner_set_acc_number(self):
+        """
```
simahawk: comment

l10n_ch_base_bank/models/invoice.py
```diff
@@ -121,15 +117,40 @@ def _check_isr(self):
                 invoice._is_isr_reference()
         return True
 
+    def write(self, vals):
+        """Override to update partner_bank_id before constraints if needed and
+        to be consistent with create
+        """
+        res = super().write(vals)
+        for rec in self:
+            if not rec.partner_bank_id and rec.type == 'out_invoice':
```
simahawk: any chance to avoid multiple write?

l10n_ch_base_bank/models/invoice.py
```diff
@@ -121,15 +117,40 @@ def _check_isr(self):
...
+                banks = rec.partner_banks_to_show()
+                if banks:
+                    rec.write({'partner_bank_id': banks[0].id})
+        return res
+
     @api.model
     def create(self, vals):
         """We override create in order to have customer invoices
```
simahawk:
```py
# make sure to have customer invoices generated by the commercial flow
# since on partner change it doesn't happen automatically
```

l10n_ch_base_bank/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,7 @@
+* Nicolas Bessi (Camptocamp)
+* Vincent Renaville <vincent.renaville@camptocamp.com>
+* Joël Grand-Guillaume <joel.grandguillaume@camptocamp.com>
+* Guewen Baconnier <guewen.baconnier@camptocamp.com>
+* Paul Catinean <paulcatinean@gmail.com>
+* Yannick Vaucher <yannick.vaucher@camptocamp.com>
+* Akim Juillerat <akim.juillerat@camptocamp.com>
```
simahawk, line 7: after this effort you can put yourself here no? :)

l10n_ch_base_bank/readme/USAGE.rst
```diff
@@ -0,0 +1,8 @@
+Account type will be discovered authomatically.
```
simahawk: *automatically

l10n_ch_base_bank/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This addons will add different bank types required by specific swiss electronic
```
simahawk: *Swiss

l10n_ch_base_bank/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
...
+payment like DTA and ISR. It allows to manage both Post and Bank systems.
+
+It'll perform some validation when entring bank account number or ESR number
```
simahawk: *entering or filling

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -12,8 +12,27 @@
 
 class TestBank(common.TransactionCase):
 
+    def setUp(self):
+        super(TestBank, self).setUp()
+        self.partner = self.env.ref('base.res_partner_12')
+        self.bank = self.env['res.bank'].create({
```
simahawk: do we need to do this for _every_ test? If not pls use SavepointCase and set tracking_disable globally

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -158,46 +174,43 @@ def test_set_ccp(self):
...
             'ccp': '10-8060-7',
         })
-        bank_acc.onchange_ccp_set_empty_acc_number()
+        bank_acc.onchange_ccp_set_acc_number()
 
         self.assertEqual(bank_acc.acc_number, '10-8060-7')
         self.assertEqual(bank_acc.bank_id, self.post_bank)
 
     def test_set_ccp_bank(self):
+        # we crate bank account
```
simahawk: *create

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -282,68 +303,79 @@ def test_name_search(self):
...
+        bank_acc._update_acc_name()
+        self.assertEqual(bank_acc.acc_number, 'CCP 46-110-7')
+        # no changes for bank changes
+        bank_acc.bank_id = self.bank
+        bank_acc._update_acc_name()
+        self.assertEqual(bank_acc.acc_number, 'CCP 46-110-7')
+        # everything cleanup
+        bank_acc.ccp = ''
+        bank_acc._update_acc_name()
+        self.assertEqual(bank_acc.acc_number, '')
```
simahawk: in all these tests I see onchanges called manually: in v12 we have a new common class to ease form behaviors testing https://www.odoo.com/documentation/12.0/reference/testing.html#odoo.tests.common.Form
might be worth to give it a try.
If you don't go for it at least drop a comment for future improvement ;)

l10n_ch_base_bank/tests/test_create_invoice.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015 Yannick Vaucher (Camptocamp SA)
+# Copyright 2012-2019 Camptocamp
```
simahawk, line 1: same story regarding SavepointCase + tracking_disable
+ pls move setup of the test class at the top (it makes no sense to have the preparation of the suite at the bottom)

l10n_ch_base_bank/tests/test_search_invoice.py
```diff
@@ -7,23 +7,31 @@
 class TestSearchInvoice(common.TransactionCase):
 
     def setUp(self):
-        super(TestSearchInvoice, self).setUp()
```
simahawk, line 10: same

l10n_ch_payment_slip/__manifest__.py
```diff
@@ -1,32 +1,36 @@
...
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-{'name': 'Switzerland - Payment Slip (BVR/ESR)',
- 'summary': 'Print ESR/BVR payment slip with your invoices',
- 'version': '10.0.1.1.0',
+{'name': 'Switzerland - ISR inpayment slip (PVR/BVR/ESR)',
+ 'summary': 'Print inpayment slip from your invoices',
+ 'version': '12.0.1.0.0',
  'author': "Camptocamp,Odoo Community Association (OCA)",
  'category': 'Localization',
  'website': 'http://www.camptocamp.com',
```
simahawk: https://github.com/OCA/l10n-switzerland

l10n_ch_payment_slip/__manifest__.py
```diff
@@ -1,32 +1,36 @@
...
      "views/bank.xml",
      "views/account_invoice.xml",
-     "wizard/bvr_batch_print.xml",
-     "wizard/bvr_import_view.xml",
+     "views/res_config_settings_views.xml",
+     "wizard/isr_batch_print.xml",
      "report/report_declaration.xml",
      "security/ir.model.access.csv"
  ],
  'demo': [],
```
simahawk: same about default/empty flags

l10n_ch_payment_slip/controllers/web.py
```diff
@@ -0,0 +1,34 @@
...
+
+from odoo.models import _
+from odoo.addons.web.controllers import main as report
+from odoo.http import content_disposition, request, route
+
+
+class ReportController(report.ReportController):
+    @route()
+    def report_routes(self, reportname, docids=None, converter=None, **data):
+        if converter == "reportlab-pdf":
```
simahawk, line 13: is reportlab still something in the modern era? :thinking:

l10n_ch_payment_slip/models/invoice.py
```diff
@@ -169,3 +171,24 @@ def _check_bvr_generatable(self):
...
+        # Delete former printed payment slip
+        ActionReport = self.env['ir.actions.report']
+        try:
+            report_payment_slip = ActionReport._get_report_from_name(
+                'l10n_ch_payment_slip.one_slip_per_page_from_invoice')
+        except IndexError:
+            report_payment_slip = False
+        if report_payment_slip and report_payment_slip.attachment:
+            for invoice in self:
+                with invoice.env.do_in_draft():
```
simahawk, line 188: Q: why in draft mode? You are not changing anything, no?

l10n_ch_payment_slip/models/payment_slip.py
```diff
@@ -27,7 +25,7 @@ class PaymentSlipSettings(object):
     """Slip report setting container"""
 
     def __init__(self, report_name, **kwargs):
-        for param, value in kwargs.iteritems():
+        for param, value in list(kwargs.items()):
```
simahawk: no need to convert to list

l10n_ch_payment_slip/report/ir_action.py
```diff
@@ -1,46 +1,170 @@
...
+        if not report:
+            report = self.search([
+                ('report_type', '=', 'reportlab-pdf'),
+                ('report_name', '=', report_name)],
+                limit=1
+            )
+        return report
+
+    @api.multi
+    def render_reportlab_pdf(self, res_ids=None, data=None):
```
simahawk, line 113: is this still used?

l10n_ch_payment_slip/static/src/js/qwebactionmanager.js
```diff
@@ -0,0 +1,37 @@
...
+    var ActionManager= require('web.ActionManager');
+    var crash_manager = require('web.crash_manager');
+    var framework = require('web.framework');
+
+
+    ActionManager.include({
+        ir_actions_report: function (action, options) {
+            var report_url = '';
+
+            if (action.report_type !== 'reportlab-pdf') {
```
simahawk, line 13: same for reportlab story

l10n_ch_payment_slip/views/report_xml_templates.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
simahawk: rename this file to `assets.xml`

l10n_ch_zip/__init__.py
```diff
@@ -1,12 +1,8 @@
 # Copyright 2011-2017 Camptocamp SA
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 from . import models
+from .hooks import import_csv_data
 
 
 def post_init(cr, registry):
```
simahawk: same as per base module: move to hooks

simahawk (changes requested):

l10n_ch_base_bank/tests/test_bank.py
i-vyshnevska: I saw it, but not in this PR, and what I saw in v12 they have improved testing for onchange by default

l10n_ch_base_bank/models/bank.py
i-vyshnevska: that is bic for switzerland post, I move it to global var at least here

l10n_ch_base_bank/models/invoice.py
i-vyshnevska: we want use here partner_banks_to_show() which comes from bank-payment repo and works only with self

l10n_ch_payment_slip/models/invoice.py
i-vyshnevska: it comes from https://github.com/OCA/l10n-switzerland/pull/411 I don't want improve it in this PR

simahawk:
@i-vyshnevska this PR is huge, it takes too long to go thru it. Can't we split the 2 modules in 2 PRs?

l10n_ch_payment_slip/models/invoice.py
simahawk: perfectly fine :+1: but please leave comments in all the places where it would be nice to rethink/refactor in the future, thanks!

i-vyshnevska:
> @i-vyshnevska this PR is huge, it takes too long to go thru it. Can't we split the 2 modules in 2 PRs?

payment_slip was partially migrated and now failing, when push base_bank, I started them separately but  to keep green merge

l10n_ch_payment_slip/static/src/js/qwebactionmanager.js
i-vyshnevska: @yvaucher  can you pls add some context about reportlab here?

yvaucher: @i-vyshnevska reportlab was chosen to generate the pdf for precision issue with printers.

We have multiple parameters to align the payment slip and it's elements vertically and horizontally. Print-out might be altered by software configuration or printer setup if not printer hardware itself.

With Qweb it was way to hard to align correctly the payment slip and the Swiss post specs would make your payment slip rejected if you don't fix in a tiny error margin.

Nowadays, it may have improved and the payment slip as we know it will soon(-ish) be replaced (starting 30th of June 2020) by the QRCode version for which the readers will be more efficient.

l10n_ch_base_bank/models/bank.py
yvaucher: I find useless to search an empty value on all fields.
But it can be improved by constructing only the domain in the `if...else` and pass the domain to the same search.

``` python
if name:
    domain = [...] + args # using name
else:
    domain = args
ids = self._search(domain, limit=limit, ...)
```

We might even use `_search` to get directly the `ids` without using `browse` implied by `search`

l10n_ch_payment_slip/static/src/js/qwebactionmanager.js
i-vyshnevska: @yvaucher I  added it's to comments for clarifying, can we pass this PR further?

l10n_ch_base_bank/models/bank.py
```diff
@@ -186,6 +214,17 @@ class ResPartnerBank(models.Model, BankCommon):
...
     )
 
+    _sql_constraints = [
+        ('isr_adherent_uniq', 'unique (isr_adherent_num, ccp)',
+         'The ISR adherent number/ccp pair must be unique !'),
+    ]
+
+    @api.model
+    def _get_supported_account_types(self):
+        rslt = super(ResPartnerBank, self)._get_supported_account_types()
```
yvaucher: Just variable naming here.

I guess `rslt` is for result please prefer `res` or `result` which are more widely used in odoo and OCA.

l10n_ch_base_bank/models/invoice.py
```diff
@@ -121,15 +117,39 @@ def _check_isr(self):
...
         type_defined = vals.get('type') or self.env.context.get('type', False)
         if type_defined == 'out_invoice' and not vals.get('partner_bank_id'):
-            user = self.env.user
-            bank_ids = user.company_id.partner_id.bank_ids
-            if bank_ids:
-                vals['partner_bank_id'] = bank_ids[0].id
+            partner = self.env.user.company_id.partner_id
+            vals['partner_bank_id'] = self._get_bank_id(
+                partner, vals.get('journal_id'), vals.get('reference_type'),
+            )
```
yvaucher: In which case can't we rely on `partner_banks_to_show`?

Can't we let `partner_bank_id` undefined unless it is manually set by the user?

One reason we could want to copy the bank on invoice is for the user to see it.
Nevertheless, I would argue that the payment mode should already allow us to invoice partners by multiple means.
And that setting the bank on the invoice should be an exception and can be done manually.

I fear that we still are making something different than the OCA standard for other localisations. I would prefer to keep the partner_bank_id empty and only refer to `partner_banks_to_show`.

We probably want to find the payment mode instead.

yvaucher (changes requested):
Only one remaining question which is about the `partner_bank_id` field on invoice.

I would like the opinion of @fclementic2c here.

l10n_ch_base_bank/models/invoice.py
i-vyshnevska: this changes we made to satisfy this constrain https://github.com/OCA/l10n-switzerland/blob/37e940e6977ef0fc63b1fce526173a4cd5cd36f5/l10n_ch_base_bank/models/invoice.py#L72
partner_banks_to_show will work on already created obj

i-vyshnevska: if we dropped constraint https://github.com/OCA/l10n-switzerland/blob/37e940e6977ef0fc63b1fce526173a4cd5cd36f5/l10n_ch_base_bank/models/invoice.py#L86
we can rid of create overriding, than only one test is failed https://github.com/OCA/l10n-switzerland/blob/37e940e6977ef0fc63b1fce526173a4cd5cd36f5/l10n_ch_base_bank/tests/test_create_invoice.py#L81
@yvaucher  maybe that test and behavior is obsolete?

i-vyshnevska: @yvaucher ping

yvaucher: How I see it:

1. As a user, I create an invoice.
2. I set the partner. (at this point the payment mode can be set by the onchange [1] )
3. I set the reference type to ISR
4. Then the constraint check if bank account (either on invoice or in payment mode) is compatible with ISR type.
   if yes -> nothing
   else -> raise error

In case of error, the user will have to change manually the payment mode or the partner_bank_id. (The error message needs to be adapted accordingly to explain that payment mode or invoice bank)

About `_get_bank_id`, it could be added to:
- the partner onchange of OCA/bank-payment/account_payment_partner
- plus a reference_type onchange.
But only if the reference type is ISR and it should also look at the payment modes.
Thus, I propose to drop it for the current PR and if the needs come we will add it in another PR.


TODO:

- Thus we can drop the part where we copy the bank account on the invoice `write` and `create`. I prefer to let the module `account_payment_partner`'s onchange do the job.
- Change the constraint to check with `partner_banks_to_show`
- Change the constraint message to reflect the possible use of either payment mode or bank field on invoice.
- Adapt the test you mentioned by calling the partner onchange or by setting the partner_bank_id (the check itself is about the constraint which is still valid)
- Drop `_get_bank_id` which could come in a further PR.


Reasons:

1. IMO we should stick to OCA/bank-payment behavior as much as possible as you could have a multi-company setup and you don't want the swiss company to mess with other country companies.
2. Less code make me more happy as long as we make it simple but not simpler :)


[1]
https://github.com/OCA/bank-payment/blob/1a21770d76ec3af9fa918b8256628257884113e9/account_payment_partner/models/account_invoice.py#L47-L49

yvaucher (changes requested):

i-vyshnevska:
@yvaucher suggested changes are made in last fixup commit, also there is improvement for test cases  by using Form class, so it's a little bit messy again , sorry

l10n_ch_base_bank/models/invoice.py
```diff
@@ -121,15 +112,22 @@ def _check_isr(self):
...
-        generated by the comercial flow as on change partner is
-        not systemtically call"""
-        type_defined = vals.get('type') or self.env.context.get('type', False)
-        if type_defined == 'out_invoice' and not vals.get('partner_bank_id'):
-            user = self.env.user
-            bank_ids = user.company_id.partner_id.bank_ids
+    def partner_banks_to_show(self):
+        """
+        Extend method from account_payment_partner to add specific
+        logic for switzerland bank payments if base method give false result
```
yvaucher: I would reword it like this, as `false` could be interpreted as `wrong`

```suggestion
        logic for switzerland bank payments if base method does not give a result
```

l10n_ch_base_bank/models/invoice.py
```diff
@@ -121,15 +112,22 @@ def _check_isr(self):
...
+        """
+        res = super().partner_banks_to_show()
+        if not res:
+            if self.journal_id:
+                return self.env['account.journal'].browse(self.journal_id.id)\
```
yvaucher: No need to rebrowse the object you already have.

```suggestion
                return self.journal_id.bank_account_id
```

yvaucher (changes requested):
Thanks for the changes.

Just spotted on comment and one line of code to improve.

yvaucher (approved):

PR #472, opened 2019-03-11
[11.0][FIX] refactor code to updates in bank-payment

i-vyshnevska:
fix mater branch due to refactoring account-payment-partner in https://github.com/OCA/bank-payment/pull/528/files some test are failing 

[UI tests](https://youtu.be/if01Ms6DcEA)

### DIFF
--- a/l10n_ch_base_bank/models/bank.py
+++ b/l10n_ch_base_bank/models/bank.py
@@ -152,8 +152,8 @@ class Bank(models.Model, BankCommon):
             ids = self.search(
                 args,
                 limit=limit
-                ).ids
-        # we sort by occurence
+            ).ids
+        # we sort by occurrence
         to_ret_ids = list(set(ids))
         to_ret_ids = sorted(
             to_ret_ids,
@@ -167,6 +167,31 @@ class ResPartnerBank(models.Model, BankCommon):
     """Inherit res.partner.bank class in order to add swiss specific fields
     and state controls
 
+    Statements:
+    acc_type could be of 3 types:
+        - postal
+        - iban
+        - bank
+
+    if account has ccp and acc_type = 'postal' we dropped acc_number and
+    compute it based on ccp, and partner
+
+    if acc_number given in 'iban' format just transform to iban format, but no
+    further modification on it, and acc_type = 'iban'
+
+    if given bank is a postal (acc_type = 'postal') and ccp != bank.ccp it's
+    mean that ccp in account is ccp of this partner in it's could be different
+    acc_number in this case recomputed by as partner_name + ccp
+
+    if given bank is a bank (acc_type = 'bank') then ccp in account should be
+    the same as bank.ccp and acc_number recomputed
+
+    if given ccp and no bank_id:
+     - check if we already have banks with the same ccp in db, if found set
+      this bank to account, update the rest
+     - if no matches this mean this is a postal type, it has no bank, we
+     set acc_number to this ccp number
+
     """
     _inherit = 'res.partner.bank'
     _compile_check_isr_add_num = re.compile('[0-9]*$')
@@ -177,7 +202,7 @@ class ResPartnerBank(models.Model, BankCommon):
         help="Your Bank adherent number to be printed "
              "in references of your ISR. "
              "This is not a postal account number."
-        )
+    )
     acc_number = fields.Char(
         string='Account/IBAN Number'
     )
@@ -185,6 +210,11 @@ class ResPartnerBank(models.Model, BankCommon):
         string='CCP/CP-Konto',
         store=True
     )
+
+    _sql_constraints = [
+        ('isr_adherent_uniq', 'unique (isr_adherent_num, ccp)',
+         'The ISR adherent number/ccp pair must be unique !'),
+    ]
 
     @api.depends('acc_number')
     def _compute_acc_type(self):
@@ -236,33 +266,52 @@ class ResPartnerBank(models.Model, BankCommon):
         return True
 
     @api.multi
-    def _get_acc_name(self):
-        """ Return an account name for a bank account
-        to use with a ccp for ISR.
-        This method makes sure to generate a unique name
-        """
-
+    def _update_acc_name(self):
+        """Check if number generated from ccp, if yes replace it on new """
         part_name = self.partner_id.name
         if not part_name and self.env.context.get('default_partner_id'):
             partner_id = self.env.context.get('default_partner_id')
             part_name = self.env['res.partner'].browse(partner_id)[0].name
+        self.acc_number = self._compute_name_ccp(part_name, self.ccp)
 
-        if part_name:
-            acc_name = _("Bank/CCP {}").format(part_name)
+    @api.multi
+    def _compute_name_ccp(self, partner_name, ccp):
+        """This method makes sure to generate a unique name"""
+        if partner_name and ccp:
+            acc_name = _("{}/CCP {}").format(partner_name, ccp)
+        elif ccp:
+            acc_name = _("CCP {}").format(ccp)
         else:
-            acc_name = _("Bank/CCP Undefined")
+            return ''
 
         exist_count = self.env['res.partner.bank'].search_count(
             [('acc_number', '=like', acc_name)])
+        # if acc_number not unique iterate on bank_accounts while not get
+        # unique number
         if exist_count:
             name_exist = exist_count
             while name_exist:
-                new_name = acc_name + " ({})".format(exist_count)
+                new_name = acc_name + " #{}".format(exist_count)
                 name_exist = self.env['res.partner.bank'].search_count(
                     [('acc_number', '=', new_name)])
                 exist_count += 1
             acc_name = new_name
         return acc_name
+
+    @api.model
+    def create(self, vals):
+        """
+        acc_number is mandatory for model, but in localization it could be not
+        mandatory when we have ccp number, so we compute acc_number in onchange
+        methods and check it here also
+        """
+        if not vals.get('acc_number') and vals.get('ccp'):
+            partner = self.env['res.partner'].browse(vals.get('partner_id'))
+            vals['acc_number'] = self._compute_name_ccp(
+                partner.name,
+                vals['ccp']
+            )
+        return super().create(vals)
 
     @api.multi
     def _get_ch_bank_from_iban(self):
@@ -273,7 +322,7 @@ class ResPartnerBank(models.Model, BankCommon):
         return clearing and self.env['res.bank'].search(
             [('clearing', '=', clearing)], limit=1)
 
-    @api.onchange('acc_number', 'acc_type')
+    @api.onchange('acc_number')
     def onchange_acc_number_set_swiss_bank(self):
         """ Set the bank when possible
         and set ccp when undefined
@@ -287,6 +336,10 @@ class ResPartnerBank(models.Model, BankCommon):
         - a bank account with CCP on the bank, we use ccp of the bank
         - otherwise there is no CCP to use
         """
+        if not self.acc_number:
+            # if account number was flashed in UI
+            self._update_acc_name()
+
         bank = self.bank_id
         ccp = False
         if self.acc_type == 'postal':
@@ -298,8 +351,6 @@ class ResPartnerBank(models.Model, BankCommon):
                 bank or
                 self.env['res.bank'].search([('bic', '=', 'POFICHBEXXX')],
                                             limit=1))
-            if not bank.is_swiss_post():
-                self.acc_number = self._get_acc_name()
         elif self.acc_type == 'iban':
             if not bank:
                 bank = self._get_ch_bank_from_iban()
@@ -308,58 +359,65 @@ class ResPartnerBank(models.Model, BankCommon):
                     ccp = self._convert_iban_to_ccp(self.acc_number.strip())
                 else:
                     ccp = bank.ccp
+
         elif self.bank_id.ccp:
             ccp = self.bank_id.ccp
         self.bank_id = bank
+
         if not self.ccp:
             self.ccp = ccp
 
     @api.onchange('ccp')
-    def onchange_ccp_set_empty_acc_number(self):
-        """ If acc_number is empty and bank ccp is defined fill it """
-        if self.bank_id:
-            if not self.acc_number and self.ccp:
-                if self.bank_id.is_swiss_post():
-                    self.acc_number = self.ccp
-                else:
-                    self.acc_number = self._get_acc_name()
+    def onchange_ccp_set_acc_number(self):
+        """If ccp changes and it's a postal bank update acc_number to ccp
+        we don't want make acc_number as computed to have possibility set it
+        manually and also avoid to shadow other logic on acc_number if exist
+       """
+        if self.acc_type == 'iban':
             return
 
+        if self.ccp:
+            if self.acc_type == 'postal':
+                # flash bank if it was previously setup, also trigger acc_type
+                # changing
+                self.bank_id = ''
+            self._update_acc_name()
+        else:
+            # flash bank if it was previously setup
+            self.acc_number = ''
+
+        if self.bank_id.is_swiss_post():
+            self.acc_number = self.ccp
+            return
+
+        # try to find appropriate bank
         ccp = self.ccp
-        if ccp and self.is_swiss_postal_num(ccp):
-            bank = (
-                self.env['res.bank'].search([('ccp', '=', ccp)], limit=1) or
-                self.env['res.bank'].search([('bic', '=', 'POFICHBEXXX')],
-                                            limit=1))
-            if not self.acc_number:
-                if not bank.is_swiss_post():
-                    self.acc_number = self._get_acc_name()
-                else:
-                    self.acc_number = self.ccp
+        if ccp and self.is_swiss_postal_num(ccp) and not self.bank_id.id:
+            bank = (self.env['res.bank'].search([('ccp', '=', ccp)], limit=1)
+                    or
+                    self.env['res.bank'].search([('bic', '=', 'POFICHBEXXX')],
+                                                limit=1))
+            if not bank.is_swiss_post():
+                self._update_acc_name()
+            else:
+                self.acc_number = self.ccp
             self.bank_id = bank
 
     @api.onchange('bank_id')
     def onchange_bank_set_acc_number(self):
-        """ If acc_number is empty and bank ccp is defined fill it """
-        if not self.bank_id:
+        """ Track bank change to update acc_name if needed"""
+        if not self.bank_id or self.acc_type == 'iban':
             return
         if self.bank_id.is_swiss_post():
-            if not self.acc_number:
-                self.acc_number = self.ccp
-            elif not self.ccp and self.is_swiss_postal_num(self.acc_number):
-                self.ccp = self.acc_number
+            self.acc_number = self.ccp
         else:
-            if not self.acc_number and self.ccp:
-                self.acc_number = self._get_acc_name()
-            elif self.acc_number and self.is_swiss_postal_num(self.acc_number):
-                self.ccp = self.acc_number
-                self.acc_number = self._get_acc_name()
+            self._update_acc_name()
 
     @api.onchange('partner_id')
     def onchange_partner_set_acc_number(self):
+        """
+        when acc_number was computed automatically we call regeneration
+        as partner name is part of acc_number
+        """
         if self.acc_type == 'bank' and self.ccp:
-            if 'Bank/CCP' in self.acc_number:
-                self.acc_number = self._get_acc_name()
-
-    _sql_constraints = [('isr_adherent_uniq', 'unique (isr_adherent_num, ccp)',
-                         'The ISR adherent number/ccp pair must be unique !')]
+            self._update_acc_name()
--- a/l10n_ch_base_bank/models/invoice.py
+++ b/l10n_ch_base_bank/models/invoice.py
@@ -121,6 +121,21 @@ class AccountInvoice(models.Model):
                 invoice._is_isr_reference()
         return True
 
+    def write(self, vals):
+        """Override to update partner_bank_id before constraints if needed and
+        to be consistent with create
+        """
+        if not self.partner_bank_id or not vals.get('partner_bank_id'):
+            type_defined = vals.get('type') or self.type
+            if type_defined == 'out_invoice':
+                partner = self.env.user.company_id.partner_id
+                journal = vals.get('journal_id') or self.journal_id.id
+                ref_type = vals.get('reference_type') or self.reference_type
+                vals['partner_bank_id'] = self._get_bank_id(
+                    partner, journal, ref_type,
+                )
+        return super().write(vals)
+
     @api.model
     def create(self, vals):
         """We override create in order to have customer invoices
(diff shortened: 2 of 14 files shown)


yvaucher (commented):
Reading https://github.com/OCA/bank-payment/pull/528/commits/95174eeaf10d61e27b3301f6fbd3a2b6ad74ed0e

I see we have now: `partner_banks_to_show()` method which should get the bank linked to a partner.
We probably want to rely on this as well.

If I get it right, the bank depends on the payment mode. Thus, this shouldn't be the role of base_bank to do this.

If we really need that, let's say for l10n_ch_payment_slip, I would rather delegate the change of behavior in l10n_ch_payment_slip. Which could be done by inheriting from `partner_banks_to_show`

Thus, to fix the tests, we might want to drop the check on bank.

i-vyshnevska:
@yvaucher  what about this, it also should be removed? https://github.com/OCA/l10n-switzerland/blob/0c98a3bbc0ce5d89ad942d87eb2635cc5677962b/l10n_ch_base_bank/models/invoice.py#L127

l10n_ch_base_bank/models/invoice.py
```diff
@@ -67,7 +67,20 @@ def onchange_reference(self):
                 self._is_isr_reference()
             except exceptions.ValidationError:
                 return
-            self.reference_type = 'isr'
+            self.write({'reference_type': 'isr'})
```
yvaucher: This change seems unnecessary and in my opinion makes it less readable.

l10n_ch_base_bank/models/invoice.py
```diff
@@ -67,7 +67,20 @@ def onchange_reference(self):
...
+
+    def _get_bank_id(self, partner, journal_id, ref_type):
+        if journal_id:
+            return self.env['account.journal'].browse(journal_id).\
+                bank_account_id.id
+        if ref_type == 'irs':
```
yvaucher:
```suggestion
        if ref_type == 'isr':
```

l10n_ch_base_bank/models/invoice.py
```diff
@@ -67,7 +67,20 @@ def onchange_reference(self):
                 self._is_isr_reference()
             except exceptions.ValidationError:
                 return
-            self.reference_type = 'isr'
+            self.write({'reference_type': 'isr'})
+
+    def _get_bank_id(self, partner, journal_id, ref_type):
```
yvaucher: Here we want to use the method to get partner either from the payment mode or on the invoice.

l10n_ch_base_bank/models/bank.py
```diff
@@ -235,34 +235,12 @@ def _check_postal_num(self):
...
-                name_exist = self.env['res.partner.bank'].search_count(
-                    [('acc_number', '=', new_name)])
-                exist_count += 1
-            acc_name = new_name
-        return acc_name
+    @api.model
+    def create(self, vals):
+        # One of the fields is mandatory
+        if not vals.get('acc_number') and vals.get('ccp'):
+            vals['acc_number'] = vals['ccp']
```
yvaucher: This will probably not work with 2 partners having the same ccp.

As we still have the constraint:

https://github.com/odoo/odoo/blob/11.0/odoo/addons/base/res/res_bank.py#L74

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -281,69 +284,3 @@ def test_name_search(self):
         self.bank.city = 'Lausanne-Centre'
         result = self.env['res.bank'].name_search('Lausanne-Centre')
         self.assertEqual(result and result[0][0], self.bank.id)
-
-    def test_multiple_isr_bank_account_for_same_partner(self):
```
yvaucher, line 285: We want to keep a similar test to check the constraint will not forbid us to do that.

Your company can have 2 accounts at the same bank for different purposes. Thus we want to be sure we can have twice the same ccp.

By the way we might be missing a test having 2 partners with same bank ccp.

yvaucher (changes requested):
I think there is a misunderstanding here.

As far as I know acc_number is still a required field and we cannot push everytime the ccp in there.
Because there is also an unicity constraint on `acc_number`.

The only thing we want here was to get rid of the direct use of `invoice.bank_id`

Each time there is a `self.bank_id` we want to use the `partner_banks_to_show` method instead.

l10n_ch_base_bank/models/invoice.py
i-vyshnevska: I could replace it only for write method, create will use old one

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -68,9 +88,9 @@ def test_bank_ccp(self):
         bank_acc.onchange_acc_number_set_swiss_bank()
 
         self.assertEqual(bank_acc.bank_id, self.bank)
-        self.assertEqual(bank_acc.acc_number, 'Bank/CCP Camptocamp')
+        self.assertEqual(bank_acc.acc_number, '46-110-7')
         self.assertEqual(bank_acc.ccp, '46-110-7')
-        self.assertEqual(bank_acc.acc_type, 'bank')
+        self.assertEqual(bank_acc.acc_type, 'postal')
```
yvaucher: Here we want to test a Bank CCP

Bank CCP != CCP

Thus we are really talking about a account type of type `bank`.

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -68,9 +88,9 @@ def test_bank_ccp(self):
         bank_acc.onchange_acc_number_set_swiss_bank()
 
         self.assertEqual(bank_acc.bank_id, self.bank)
-        self.assertEqual(bank_acc.acc_number, 'Bank/CCP Camptocamp')
+        self.assertEqual(bank_acc.acc_number, '46-110-7')
```
yvaucher: This is not correct.

We are expecting to get `Camptocamp/CCP 46-110-7`

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -80,27 +100,35 @@ def test_bank_ccp_no_found(self):
...
+        # if acc_number given by user don't update it
         self.assertEqual(bank_acc.acc_number, '10-8060-7')
         self.assertEqual(bank_acc.ccp, '10-8060-7')
         self.assertEqual(bank_acc.acc_type, 'postal')
 
         # specify that it isn't a postal account
+        self.bank.write({
+            'cpp': '46-110-7',
+            'bic': 'POFICHBEXXX'
+        })
```
yvaucher: In setup you already have 
``` python
        self.bank = self.env['res.bank'].create({
            'name': 'Alternative Bank Schweiz AG',
            'bic': 'ALSWCH21XXX',
            'clearing': '38815',
            'ccp': '46-110-7',
        })
        self.post_bank = self.env['res.bank'].search(
            [('bic', '=', 'POFICHBEXXX')])
        if not self.post_bank:
            self.post_bank = self.env['res.bank'].create({
                'name': 'PostFinance AG',
                'bic': 'POFICHBEXXX',
                'clearing': '9000',
                'ccp': '10-8060-7',
            })
```

Please don't mixup those.

post_bank customers have a direct CCP account (it is unique only one bic exists for the swiss post "Postfinance")
bank customers use the CCP of their bank.

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -152,52 +180,15 @@ def test_other_bank(self):
...
-        bank_acc = self.env['res.partner.bank'].new({
-            'partner_id': self.partner.id,
-            'acc_number': None,
-            'bank_id': self.bank.id,
-            'ccp': '10-8060-7'
-        })
-        bank_acc.onchange_ccp_set_empty_acc_number()
-
-        self.assertEqual(bank_acc.acc_number, 'Bank/CCP Camptocamp')
-        self.assertEqual(bank_acc.bank_id, self.bank)
```
yvaucher, line 176: Why are those tests removed?

l10n_ch_base_bank/models/bank.py
```diff
@@ -179,19 +179,34 @@ class ResPartnerBank(models.Model, BankCommon):
...
+         "CHECK( (ccp IS NOT NULL) or (acc_number IS NOT NULL))",
+         'Number is mandatory please setup CCP or Account number'),
+    ]
+
     @api.depends('acc_number')
     def _compute_acc_type(self):
         todo = self.env['res.partner.bank']
         for rec in self:
+            # try to find ccp in acount number and set bank from it
+            names = rec.acc_number.split('CCP-', 1)
```
yvaucher: We don't want to do this check.

`postal` type must be used only for account for which the bank is Postfinance. Bank's CCP are considered as 'bank' type. This has an impact on payment slip.

l10n_ch_base_bank/models/bank.py
```diff
@@ -236,34 +251,53 @@ def _check_postal_num(self):
...
+        elif not part_name:
+            self.acc_number = _("Undefined")
 
-        if part_name:
-            acc_name = _("Bank/CCP {}").format(part_name)
+    @api.multi
+    def _compute_name_ccp(self, partner_name, ccp):
+        """This method makes sure to generate a unique name"""
+        if partner_name:
+            acc_name = _("{}/CCP-{}").format(partner_name, ccp)
```
yvaucher: Please remove the dash. CCP is not part of the number.
```suggestion
            acc_name = _("{}/CCP {}").format(partner_name, ccp)
```

l10n_ch_base_bank/models/bank.py
```diff
@@ -273,7 +307,7 @@ def _get_ch_bank_from_iban(self):
         return clearing and self.env['res.bank'].search(
             [('clearing', '=', clearing)], limit=1)
 
-    @api.onchange('acc_number', 'acc_type')
+    @api.onchange('acc_number', 'acc_type', 'ccp')
```
yvaucher: This onchange is there to seach and set the bank if possible plus the `ccp` depending on `acc_number` it will break the UI to add `ccp` in it.

We test if `self.ccp` is set only to know if the `cpp` was already set. In which case we don't want to update it.

l10n_ch_base_bank/models/bank.py
```diff
@@ -298,8 +332,6 @@ def onchange_acc_number_set_swiss_bank(self):
                 bank or
                 self.env['res.bank'].search([('bic', '=', 'POFICHBEXXX')],
                                             limit=1))
-            if not bank.is_swiss_post():
-                self.acc_number = self._get_acc_name()
```
yvaucher, line 302: This is needed. It is a bit counter intuitive, because at this point, `acc_type` is `postal` do to the entry of the user in `acc_number` which updated `acc_type`. And after we have changed the `acc_number` it might become `bank`.

For instance how it was working in previous versions:

1. Fill `acc_number` with 10-8060-7
2. Onchange on `acc_number` changes `acc_type` to `postal`
3. Onchange on `acc_number` + `acc_type` starts
  3a. on detection of a 'postal' account type search for a bank that has the CCP.
  3b. It matches with BAS bank
  3c. change `acc_number` to a generated name.
  3d. set `bank_id`
  3e. copy the `ccp` that was entered in `acc_number` in `ccp` field.
4. As `acc_number` changed this should trigger onchange on `acc_number` again to set `acc_type` = `bank`

l10n_ch_base_bank/models/bank.py
```diff
@@ -315,51 +347,41 @@ def onchange_acc_number_set_swiss_bank(self):
             self.ccp = ccp
 
     @api.onchange('ccp')
-    def onchange_ccp_set_empty_acc_number(self):
-        """ If acc_number is empty and bank ccp is defined fill it """
+    def onchange_ccp_set_acc_number(self):
+        """If ccp changing and it's a postal bank update acc_number to ccp
```
yvaucher:
```suggestion
        """If ccp changes and it's a postal bank update acc_number to ccp
```

l10n_ch_base_bank/models/bank.py
```diff
@@ -315,51 +347,41 @@ def onchange_acc_number_set_swiss_bank(self):
...
+        we don't want make acc_number as compute to have possibility set it
```
yvaucher:
```suggestion
        we don't want make acc_number as computed to have possibility to set it
```

l10n_ch_base_bank/models/bank.py
```diff
@@ -315,51 +347,41 @@ def onchange_acc_number_set_swiss_bank(self):
...
+        manually and also avoid to shadow other logic on acc_number if exist
+        """
         if self.bank_id:
-            if not self.acc_number and self.ccp:
```
yvaucher, line 321: Here we are in an onchange.

We cannot assume the field is set nor is empty.
This check was there to avoid an overwrite of what the user would have written in `acc_number` field.

I think it would be annoying for the user to enter the acc_number field, then edit the ccp field and see his first edition overwritten.

i-vyshnevska:
@yvaucher  can you return to review, it's finally green, I have one error from but this is just not properly muted logger

yvaucher (changes requested):
This is more complicated that it might looks like. The onchange have a sequence required to work correctly from the view. If we were in version 12.0 I would suggest to use the new unit test class `Form` in order to make proper onchange testing.

Please give me more insights about why you do all those changes.

The tests were on purpuse done with `new` as it was mainly to test the onchanges.

So I realise that the `create()` won't work. I don't know why it is needed to change them. But, here you want to test in a onchange context. You might be able to fix that by using an context manager for onchange.

```
with self.env.do_in_onchange()
```

Plus do some manual testing in `Contacts -> Bank Accounts -> Bank Accounts`

i-vyshnevska:
new method avoid constrains in db, or some behavior are not clear,  you can see this on instance of acc_number earlier tests were successfully  running when it was empty which is wrong as core make it required

l10n_ch_base_bank/tests/test_bank.py
i-vyshnevska: this change also have come from replacement  new -> create
we compute acc_type as postal here https://github.com/OCA/l10n-switzerland/blob/0c98a3bbc0ce5d89ad942d87eb2635cc5677962b/l10n_ch_base_bank/models/bank.py#L190
and based on this setup bank
https://github.com/OCA/l10n-switzerland/blob/0c98a3bbc0ce5d89ad942d87eb2635cc5677962b/l10n_ch_base_bank/models/bank.py#L292

l10n_ch_base_bank/tests/test_bank.py
i-vyshnevska: my logic is that if number is given from form UI we don't change it, if need do it so it rather make compute via name_get  like partner+acc_number+ccp

l10n_ch_base_bank/tests/test_bank.py
i-vyshnevska: ok

l10n_ch_base_bank/models/bank.py
i-vyshnevska: this change have come from test https://github.com/OCA/l10n-switzerland/blob/0c98a3bbc0ce5d89ad942d87eb2635cc5677962b/l10n_ch_base_bank/tests/test_bank.py#L285
this test also start failing as acc_number should be unique and only ccp can be the same for few partners so old test probably are not valid

i-vyshnevska: about UI I've forgot, sorry will check

l10n_ch_base_bank/models/bank.py
i-vyshnevska: ok, than we would have obsolete data here https://github.com/i-vyshnevska/l10n-switzerland/blob/b38e620a876b8577eac326450a5af294daa8376c/l10n_ch_base_bank/tests/test_bank.py#L250 
we would have 'Camptocamp/CCP 46-110-7', ccp part wouldn't be changed
I think we should move some logic to write and give user not so complicated interface, maybe set domain on bank_id etc

yvaucher:
> new method avoid constrains in db, or some behavior are not clear, you can see this on instance of acc_number earlier tests were successfully running when it was empty which is wrong as core make it required

Ok, thanks for the clarification.
To be noted, the incomplete filling of the vals in `new(vals)` was on purpose. That way calling onchange was done with only the value we wanted in entry. Thus we really wants to restore those test with the same aim which can be done with  `with self.env.do_in_onchange()`

here is an example:
https://github.com/odoo/odoo/blob/12.0/odoo/addons/test_new_api/tests/test_new_fields.py#L958

l10n_ch_base_bank/tests/test_bank.py
i-vyshnevska: I've dropped empty acc_number logic, and other part is tested in other tests,
I think we should split testing for UI  probably with new() whatever, and actually testing for create, write

l10n_ch_base_bank/models/bank.py
```diff
@@ -179,19 +179,34 @@ class ResPartnerBank(models.Model, BankCommon):
...
         string='CCP/CP-Konto',
         store=True
     )
 
+    _sql_constraints = [
+        ('isr_adherent_uniq', 'unique (isr_adherent_num, ccp)',
+         'The ISR adherent number/ccp pair must be unique !'),
+        ('check_numbers',
+         "CHECK( (ccp IS NOT NULL) or (acc_number IS NOT NULL))",
+         'Number is mandatory please setup CCP or Account number'),
```
yvaucher: This constraint will always be true as acc_number is required.

i-vyshnevska:
@yvaucher changes have been made + in description UI interaction of the form

l10n_ch_base_bank/models/bank.py
```diff
@@ -308,58 +359,69 @@ def onchange_acc_number_set_swiss_bank(self):
...
+            return
+
+        if self.ccp:
+            if self.acc_type == 'postal':
+                # flash bank if it previously setup also it trigger acc_type
+                # changing
+                self.bank_id = ''
+            self._update_acc_name()
+        else:
+            # flash bank if it previously setup
```
yvaucher:
```suggestion
            # flash bank if it was previously setup
```

l10n_ch_base_bank/models/bank.py
```diff
@@ -308,58 +359,69 @@ def onchange_acc_number_set_swiss_bank(self):
...
+        """If ccp changes and it's a postal bank update acc_number to ccp
+        we don't want make acc_number as computed to have possibility set it
+        manually and also avoid to shadow other logic on acc_number if exist
+       """
+        if self.acc_type == 'iban':
+            return
+
+        if self.ccp:
+            if self.acc_type == 'postal':
+                # flash bank if it previously setup also it trigger acc_type
```
yvaucher:
```suggestion
                # flash bank if it was previously setup, also trigger acc_type
```

l10n_ch_base_bank/models/bank.py
```diff
@@ -308,58 +359,69 @@ def onchange_acc_number_set_swiss_bank(self):
...
+                                                limit=1))
+            if not bank.is_swiss_post():
+                self._update_acc_name()
+            else:
+                self.acc_number = self.ccp
             self.bank_id = bank
 
     @api.onchange('bank_id')
     def onchange_bank_set_acc_number(self):
         """ If acc_number is empty and bank ccp is defined fill it """
```
yvaucher: Here the docstring doesn't match anymore.

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -61,14 +80,18 @@ def test_bank_iban_foreign(self):
...
             'acc_number': '46-110-7',
         })
+
         bank_acc.onchange_acc_number_set_swiss_bank()
+        # ccp also changed
+        bank_acc.onchange_ccp_set_acc_number()
 
-        self.assertEqual(bank_acc.bank_id, self.bank)
-        self.assertEqual(bank_acc.acc_number, 'Bank/CCP Camptocamp')
+        self.assertFalse(bank_acc.bank_id.id)
```
yvaucher: here shouldn't we find still the bank with it's ccp number?

Onchange ccp should have set the field `bank_id` with 
``` python
self.env['res.bank'].search([('ccp', '=', ccp)], limit=1)
```

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -235,38 +243,47 @@ def test_get_account_number(self):
         acc_num = bank_acc.get_account_number()
         self.assertEqual(acc_num, bank_acc.ccp)
 
-    def test_onchange_bank_empty_acc_number(self):
-        bank_acc = self.env['res.partner.bank'].new({
+    def test_onchange_bank_default_acc_number(self):
```
yvaucher:
```suggestion
    def test_create_bank_default_acc_number(self):
```

l10n_ch_base_bank/tests/test_bank.py
```diff
@@ -158,46 +174,38 @@ def test_set_ccp(self):
...
+            # in result we should get new ccp number as we have bank_id and
+            # this he has ccp, new acc_number
 
-        self.assertEqual(bank_acc.acc_number, 'Bank/CCP Camptocamp')
-        self.assertEqual(bank_acc.bank_id, self.bank)
+            self.assertEqual(bank_acc.acc_number, 'Camptocamp/CCP 10-8060-7')
+            self.assertEqual(bank_acc.ccp, '10-8060-7')
+            self.assertEqual(bank_acc.bank_id, self.bank)
 
     def test_set_ccp_post(self):
```
yvaucher:
```suggestion
    def test_create_ccp_post(self):
```

We might still want both the create and the onchange tests.

yvaucher (changes requested):
Thanks

Looks almost good. I added few remarks, with mostly nitpickings.

Plus there is some lint to do to make travis happy:

```
l10n_ch_base_bank/models/bank.py:190:80: E501 line too long (80 > 79 characters)
l10n_ch_base_bank/models/bank.py:397:21: E126 continuation line over-indented for hanging indent
l10n_ch_base_bank/models/bank.py:397:80: E501 line too long (80 > 79 characters)
l10n_ch_base_bank/models/invoice.py:145:21: E126 continuation line over-indented for hanging indent
l10n_ch_base_bank/models/invoice.py:145:80: E501 line too long (80 > 79 characters)
l10n_ch_base_bank/models/invoice.py:146:17: E121 continuation line under-indented for hanging indent
l10n_ch_base_bank/tests/test_bank.py:288:5: E303 too many blank lines (2)
l10n_ch_base_bank/tests/test_create_invoice.py:37:80: E501 line too long (83 > 79 characters)
```

i-vyshnevska:
@yvaucher comments got accounted, build green

l10n_ch_base_bank/models/bank.py
```diff
@@ -236,34 +266,53 @@ def _check_postal_num(self):
...
 
-        if part_name:
-            acc_name = _("Bank/CCP {}").format(part_name)
+    @api.multi
+    def _compute_name_ccp(self, partner_name, ccp):
+        """This method makes sure to generate a unique name"""
+        if partner_name and ccp:
+            acc_name = _("{}/CCP {}").format(partner_name, ccp)
+        elif partner_name:
+            acc_name = _("{}/CCP").format(partner_name)
```
yvaucher: When emptying acc_number or ccp the acc_number field is again filled.

If we don't have a ccp number it is better to keep it empty the user might not be filling a CCP account.


Thus I would suggest to replace it by:

``` python
        if partner_name and ccp:
            acc_name = _("{}/CCP {}").format(partner_name, ccp)
        elif ccp:
            acc_name = _("CCP {}").format(ccp)
        else:
            return ''
```

yvaucher (changes requested):
@i-vyshnevska Thanks I tested in locally and it works as expected.

Just one last thing to fix. When having empty ccp or empty acc_number we have that auto filling that could drive a user crazy. I suggest to remove the auto filling when ccp is not defined.

yvaucher (approved):

PR #475, opened 2019-03-19
[10.0][FIX] Bypass unsupported address nodes in CH SEPA pain

ecino:
This will fix CH Pain files validation. Depends on https://github.com/OCA/bank-payment/pull/569 refactor.

PR #476, opened 2019-03-19
[10.0] l10n_ch_pain_base: Fix C-Level address block not valid

### LINKED ISSUE #461: [10.0] pain001 generated file is not valid.
The pain file generated is not valid with six documentation

the generated file has AdrLine and PstCd and when i upload it to postfinance test platrform i get the error. 

`The structure of the XML element /Document/CstmrCdtTrfInitn/PmtInf[1]/CdtTrfTxInf[1]/Cdtr/PstlAdr is not valid.
    Error message:
        Structured address data should not be used together with unstructured data.
    Error code:
        CH16 (ElementContentFormallyIncorrect)
    File position:
        Line:   58
        Column: 19`

The documentation says.

`If possible structured.
Simultaneous use of the structured elements <StrtNm>/<BldgNb>/
<PstCd>/<TwnNm> and the unstructured element <AdrLine> is not
permitted.`

as seen in generate_party_block -> account_payment_order.py the code will generate AdrLine and PstCd, TwnNm it those fields are present

```
if partner.country_id:
    postal_address = etree.SubElement(party, 'PstlAdr')
    if gen_args.get('pain_flavor').startswith(
            'pain.001.001.') or gen_args.get('pain_flavor').startswith(
            'pain.008.001.'):
        if partner.zip:
            pstcd = etree.SubElement(postal_address, 'PstCd')
            pstcd.text = self._prepare_field(
                'Postal Code', 'partner.zip',
                {'partner': partner}, 16, gen_args=gen_args)
        if partner.city:
            twnnm = etree.SubElement(postal_address, 'TwnNm')
            twnnm.text = self._prepare_field(
                'Town Name', 'partner.city',
                {'partner': partner}, 35, gen_args=gen_args)
    country = etree.SubElement(postal_address, 'Ctry')
    country.text = self._prepare_field(
        'Country', 'partner.country_id.code',
        {'partner': partner}, 2, gen_args=gen_args)
    if partner.street:
        adrline1 = etree.SubElement(postal_address, 'AdrLine')
        adrline1.text = self._prepare_field(
            'Adress Line1', 'partner.street',
            {'partner': partner}, 70, gen_args=gen_args)
```

So my question is, how to i go about overwriting this, best place is to overwrite it in l10n_ch_pain_credit_transfer in generate_party_block -> account_payment_order.py  ?

Im new to odoo so sorry if there is somewhere answer to this question.

p-tombez:
Fixes #461

### DIFF
--- a/l10n_ch_pain_base/__manifest__.py
+++ b/l10n_ch_pain_base/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Switzerland - ISO 20022",
     "summary": "ISO 20022 base module for Switzerland",
-    "version": "10.0.1.0.2",
+    "version": "10.0.1.0.3",
     "category": "Finance",
     "author": "Akretion,Camptocamp,Odoo Community Association (OCA)",
     "license": "AGPL-3",
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -105,3 +105,47 @@ class AccountPaymentOrder(models.Model):
             return super(AccountPaymentOrder, self).generate_party_acc_number(
                 parent_node, party_type, order, partner_bank, gen_args,
                 bank_line=bank_line)
+
+    @api.model
+    def generate_party_block(
+            self, parent_node, party_type, order, partner_bank, gen_args,
+            bank_line=None):
+        res = super(AccountPaymentOrder, self).generate_party_block(
+            parent_node, party_type, order, partner_bank, gen_args, bank_line
+        )
+
+        # Convert address to unstructured mode for SEPA compliance
+        for addr in parent_node.xpath('.//PstlAdr'):
+            loc = []
+            zip_node = addr.xpath('.//PstCd')
+            if zip_node:
+                loc.append(zip_node[0].text)
+            city_node = addr.xpath('.//TwnNm')
+            if city_node:
+                loc.append(city_node[0].text)
+
+            strt = []
+            strt_node = addr.xpath('.//StrtNm')
+            if strt_node:
+                strt.append(strt_node[0].text)
+            bldg_node = addr.xpath('.//BldgNb')
+            if bldg_node:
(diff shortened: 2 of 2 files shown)


ecino:
This is another way of doing the fix I proposed with #475. Not sure this one is better, as you change back the generated XML afterwards instead of changing how it's generated.

p-tombez:
@ecino Indeed, but this also fixes the address generation as it's wrong according to the specification : https://www.six-group.com/interbank-clearing/dam/downloads/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf (page 58-59). First AdrLine occurrence must contain street name + number and second occurrence zip code + city name, not street and street2 as it is currently.
So you should also take that in consideration in your proposed change

yvaucher:
@ecino If I get it right, #475 removes entirely the address while this PR move in in unstructured fields.

It seems the current solution is better. But, could you give it a review as you worked on the same issue?

Relevant info in the guideline is indeed:

> In  the  case  of  SEPA  pay-ments, it is recommended that the "Country" and "Address Line" elements are used.

yvaucher:
@p-tombez wouldn't it be better to refactor 

https://github.com/OCA/bank-payment/blob/9a7cb4f62109249a24a797b5d03b47c865a63acb/account_banking_pain_base/models/account_payment_order.py#L383

By extracting the address part in method. Thus, make it possible with an option to use structured or unstructured ? Option that we would force on l10n_ch_pain_base.

p-tombez:
@yvaucher The fact is that the ISO-20022 allows the usage of structured and unstructured blocks (thus easing the way the addresses are handled, e.g. no need to extract the building number from the street address). It's only the swiss specification that disallow the usage of mixed content...
Maybe we could isolate the address part in a method and override it in l10n-switzerland

p-tombez:
I'll made a new version with a refactor of account_banking_pain_base

ecino:
I agree that refactoring account_banking_pain_base is the way to go. I'd be happy to review your next PR.

PR #479, opened 2019-03-25
[MIG] l10n_ch_bank: Migration to 12.0

TDu:


yvaucher (approved):

Martronic-SA (approved):
Doesn't work with recent changes in l10n_ch_base_bank, see below.

Martronic-SA (commented):
ccp field is no more present in l10n_ch_base_bank since https://github.com/OCA/l10n-switzerland/pull/509
So installation of this PR throws an error

yvaucher:
I force pushed fixes from @grindtildeath to replace `ccp` by `l10n_ch_postal` as done in `l10n_ch_base_bank`

yvaucher:
As explained in #509 the drop of ccp field from res.bank follows the logic that this field is not really used. Suppliers always provide an account number ether an IBAN or a postal account.

I also want to remind that on May 2020 the ISR will be replaced by Swiss QR Invoices. Which means only IBAN will be used. Your suppliers will provide you unique IBAN or QR-IBAN, which will be unique numbers. Which means, it won't be necessary to have a separate field from `acc_number`. So in the future we might even remove the `l10n_ch_postal` field on res.partner.bank.

grindtildeath:
@TDu @yvaucher I messed this PR when rebasing it, as I renamed the column in `res.bank.csv` instead of dropping it. Please merge https://github.com/TDu/l10n-switzerland/pull/2

Martronic-SA:
@yvaucher @grindtildeath It's ok now with ccp / l10n_ch_postal column removed from csv file.

yvaucher:
@TDu @grindtildeath there is an error with unit test `test_iban_postal_with_spaces`

Martronic-SA:
@TDu @grindtildeath @yvaucher 
The errors seems to be related to the Postfinance Clearing number which is imported as '9000' but the l10n_ch_base_bank searches postfinance bank with clearing number = '09000' (CH_POSTFINANCE_CLEARING = "09000")
What is wrong there? the clearing number in csv file should be 09000?

yvaucher:
@Martronic-SA I would guess that there is a integer conversion that makes us look the leading `0`. 

I checked in older data, we had already 9000 instead of 09000 in the xml file before we converted it to a csv file, doesn't mean it's right, but we might just want to complete the clearing by filling with 0's on it's left.

https://github.com/OCA/l10n-switzerland/blob/7fd56dedfec7db0d629f9d69c3b471ddb023b104/l10n_ch_bank/data/bank.xml

yvaucher:
In the source CSV (with tabs) file:

https://www.six-group.com/interbank-clearing/dam/downloads/bc-bank-master/bcbankenstamm

We do have 09000, thus I think our curent file is wrong.

Martronic-SA:
> 
> 
> In the source CSV (with tabs) file:
> 
> https://www.six-group.com/interbank-clearing/dam/downloads/bc-bank-master/bcbankenstamm
> 
> We do have 09000, thus I think our curent file is wrong.

In fact no, the clearing in the file is "9000", you have 099000 which means 09 as group nad then 9000 as BC number (clearing).
Maybe the clearing (whish should be 5 digits according to the directive here:
[https://www.six-group.com/interbank-clearing/dam/downloads/en/bc_bank_master/bc_records.pdf](https://www.six-group.com/interbank-clearing/dam/downloads/en/bc_bank_master/bc_records.pdf)
should be completed by "zero" on the left to be 5 digits?

Martronic-SA:
I think this is more the "_get_ch_bank_from_iban(" method of l10n_ch_base_bank that should remove the  leading "0" before search for the bank. the clearing number can be 3 to 5 digits.
Maybe Clearing number shoud be forced not to accept leading "0"

yvaucher:
@Martronic-SA You are right.

According to this, the clearing number is padded with zeros on the left. Thus it make sense to trim those zeros when doing a match.
http://www.swissiban.com/

Which means `09000` is the format in the iban and `9000` is the real clearing number.


@TDu can you fix this?

TDu:
@Martronic-SA and @yvaucher 
I did a fix is that what was expected ? Also Travis is green now.

Martronic-SA:
> 
> 
> @Martronic-SA and @yvaucher
> I did a fix is that what was expected ? Also Travis is green now.

Seems ok for me, but as long as the user doesn't enter the bank clearing number as "09000".
Shoud it be covered here? like a first search with exact clearing number match ("9000") and if nothing is found, a second search with leading '0' ("09000")?

yvaucher:
@Martronic-SA We should consider entering 09000 as wrong. It's only an acceptable format inside the form view.

KISS: usually a user will get his clearing code from the web page or some documents of the bank. And we can assume it will be correct there.
Furthermore, with the module `l10n_ch_bank` that contains already the clearings it should be fine in 99% of the cases.

PR #482, opened 2019-03-28
[10.0] l10n_ch_pain_base: Fix empty address block generation

p-tombez:
In the Swiss specifications, the block `AdrLine` must not be empty, so we do not include it if the partner does not have `street` and `street2` set.
Also, as per the specs, the first occurrence of this block has to be the address and the second occurrence the zip code + city. If the first is not included (no street and street2), the zip code and city must also not be included.


gurneyalex (approved):
an additional unit test would be nice, but the fix looks fine to me.

i-vyshnevska (approved):
LG tests fails not related, do we need populate this fix on v11?

PR #484, opened 2019-04-18
[11.0][IMP] updates for partenr_banks_to_show method

i-vyshnevska:
improve some logic on invoicing and bank searching


l10n_ch_base_bank/models/bank.py
```diff
@@ -369,10 +378,9 @@ def onchange_acc_number_set_swiss_bank(self):
 
     @api.onchange('ccp')
     def onchange_ccp_set_acc_number(self):
-        """If ccp changes and it's a postal bank update acc_number to ccp
-        we don't want make acc_number as computed to have possibility set it
-        manually and also avoid to shadow other logic on acc_number if exist
-       """
+        # If ccp changes and it's a postal bank update acc_number to ccp
+        # we don't want make acc_number as computed to have possibility set it
+        # manually and also avoid to shadow other logic on acc_number if exist
```
yvaucher: Why are the docstring replaced by comments?

l10n_ch_base_bank/tests/test_search_invoice.py
```diff
@@ -29,14 +29,15 @@ def setUp(self):
             'type': 'bank',
             'code': 'BNK42',
             'bank_id': bank.id,
-            'bank_acc_number': '01-1234-1',
+            'bank_acc_number': '10-8060-7',
```
yvaucher: Why do this value change?

yvaucher (changes requested):
2 questions

Plus I spotted a typo in the commit message `partenr`

l10n_ch_base_bank/models/invoice.py
```diff
@@ -1,7 +1,6 @@
-# Copyright 2012 Nicolas Bessi (Camptocamp SA)
-# Copyright 2015 Yannick Vaucher (Camptocamp SA)
+# Copyright 2012-2019 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import models, api, _
+from odoo import models, api, fields, _
```
yvaucher: you can remove `fields` from the imports

l10n_ch_base_bank/tests/test_search_invoice.py
i-vyshnevska: bank_acc_number should be unique, when I'd relapsed on create it's start failing from which point I don't remeber

i-vyshnevska: ok, I restored the initials numbers (before my fix)

yvaucher (approved):

PR #485, opened 2019-04-18
[12.0][FIX] drop never running logic, cleanup

i-vyshnevska:
fast track updates to https://github.com/OCA/l10n-switzerland/pull/470


yvaucher (approved):
Approving and fast-tracking it.

PR #487, opened 2019-04-22
[11.0][FIX] l10n_ch_pain_base: incorrect compute_sepa_final_hook

BT-tpiric:
Removed incorrect compute_sepa_final_hook from AccountPaymentOrder.
pain_flavor is the used xsd version for pain files, for example pain.001.001.03.ch.02.
If the "ch" is in pain_flavor, it doesn't mean that the used payment type is "ISR".

The pain.001.001.03.ch.02 definition covers all current possible payment types in Switzerland (national, cross-border, SEPA, etc.), see https://www.six-group.com/interbank-clearing/dam/downloads/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf, sites 18 to 21.

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -10,16 +10,6 @@ class AccountPaymentOrder(models.Model):
     _inherit = 'account.payment.order'
 
     @api.multi
-    def compute_sepa_final_hook(self, sepa):
-        self.ensure_one()
-        sepa = super().compute_sepa_final_hook(sepa)
-        pain_flavor = self.payment_mode_id.payment_method_id.pain_version
-        # ISR orders cannot be SEPA orders
-        if pain_flavor and '.ch.' in pain_flavor:
-            sepa = False
-        return sepa
-
-    @api.multi
     def generate_pain_nsmap(self):
         self.ensure_one()
         nsmap = super().generate_pain_nsmap()


yvaucher:
Use of ISR is defined by the use of local_instrument `CH01` (`Type 1`)

The comment might be wrong but we still don't want to trigger the addition of
`SvcLvl/Cd/SEPA` in our XML files unless we have a `Type 5`.

This must be used with `Type 5` for other I don't think we want to add it. IIRC some bank refuse and other ignore `SvcLvl` for `Type 1`

Ref  https://www.six-group.com/interbank-clearing/dam/downloads/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf Pages 27 lines 2.8 and 2.9

EDIT: And the table at page 19 mention it must not be used for other types than `Type 5`

l10n_ch_pain_base/models/account_payment_order.py
```diff
@@ -9,16 +9,6 @@
...
-    @api.multi
-    def compute_sepa_final_hook(self, sepa):
-        self.ensure_one()
-        sepa = super().compute_sepa_final_hook(sepa)
-        pain_flavor = self.payment_mode_id.payment_method_id.pain_version
-        # ISR orders cannot be SEPA orders
-        if pain_flavor and '.ch.' in pain_flavor:
-            sepa = False
-        return sepa
-
```
yvaucher, line 21: This will have a side effect for `Type 3`

yvaucher (changes requested):
This doesn't solve the whole issue.

`account_banking_pain_base` module `compute_sepa` checks the acc_type and the currency.

In case of an iban with eur it should use the `Type 3` and in that case `SEPA` must not be set.
I don't think we should use Foreign SEPA for internal transfer of `EUR` currency.

To fix it we need to check:

``` python
            
            # if sepa is true acc_type == iban and currency == EUR
            # this defines a case of internal transfer in Switzerland (Type 3 with IBAN and EUR)
            for pline in order.payment_line_ids:
                if sepa and pline.partner_bank_id.acc_number.startswith('CH'):
                    sepa = False
                    break
```

yvaucher:
I'll close this in favor of https://github.com/OCA/l10n-switzerland/pull/487

PR #491, opened 2019-05-03
[FIX] fix test_postfinance_xml_import error

i-vyshnevska:
error appeared in test_postfinance_xml_import from some odoo core improvements


l10n_ch_bank_statement_import_postfinance/tests/test_postfinance_xml.py
```diff
@@ -108,6 +108,9 @@ def test_attachement_extraction(self):
 
 class PostFinanceImportTest(common.TransactionCase):
 
+    at_install = False
```
umiphos, line 111: Reading your code I think you want to do this:
```python
from odoo.tests import common
@common.at_install(False)
@common.at_install(False)

class PostFinanceImportTest(common.TransactionCase):
```

That way you say that this test will be executed at the end of the installation of all the dependencies.

PR #498, opened 2019-06-12
[10.0][postfinance-import] Fix bug of tiff to png conversion, by removing dependency to Wand library

ecino:
- Use PIL instead of ImageMagick which has a bug for some tiff files


yvaucher (approved):
LGTM

PR #499, opened 2019-07-08
 [12.0][ADD] ebill_paynet module 

TDu:
ToDo
 - [ ] Password on paynet.service need to be handled differently
 - [ ] Test, implement B2C message generation


ebill_paynet/tests/test_invoice_message.py
```diff
@@ -0,0 +1,160 @@
...
+from xmlunittest import XmlTestMixin
+
+from .common import compare_xml_line_by_line
+
+
+@freeze_time("2019-06-07 09:06:00")
+class TestInvoiceMessage(SingleTransactionCase, XmlTestMixin):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
i-vyshnevska, line 19: cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))

i-vyshnevska (commented):
I don't see reason why test are failing, the only reason can be it's call to "_onchange_partner_id"
triggered by transmit_method_id, 
you can try to add post_istall = False to l10n_ch_base_bank tests

ebill_paynet/components/api.py
```diff
@@ -0,0 +1,51 @@
+# import base64
```
grindtildeath: Please add header

ebill_paynet/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,14 @@
+## Create a service
+
+To create a service you need to be registred with SIXT Paynet service. Then the configuration of the service can be done in `Accounting - Configuration - Payments - Paynet Service`
+
+## Configure the customers
+
+A customer that wants to receives his invoices through Paynet will also need to register trhough the service.
```
grindtildeath:
```suggestion
A customer that wants to receive his invoices through Paynet will also need to register through the service.
```

grindtildeath (approved):
LGTM :+1:

ebill_paynet/components/api.py
```diff
@@ -0,0 +1,51 @@
...
+WSDL_DOC = os.path.join(os.path.dirname(__file__), 'wsdl', 'DWSPayNet.wsdl')
+SSL_PROD_CERTIFICATE = os.path.join(
+    os.path.dirname(__file__), 'certificats', 'prod_services_chain.pem'
+)
+SSL_TEST_CERTIFICATE = os.path.join(
+    os.path.dirname(__file__), 'certificats', 'test_services_chain.pem'
+)
+
+
+class PayNetDWS(object):
```
sebalix: No need to inherit from `object` with Python 3.

ebill_paynet/components/api.py
```diff
@@ -0,0 +1,51 @@
...
+from zeep import Client, Settings
+from zeep.transports import Transport
+
+WSDL_DOC = os.path.join(os.path.dirname(__file__), 'wsdl', 'DWSPayNet.wsdl')
+SSL_PROD_CERTIFICATE = os.path.join(
+    os.path.dirname(__file__), 'certificats', 'prod_services_chain.pem'
+)
+SSL_TEST_CERTIFICATE = os.path.join(
+    os.path.dirname(__file__), 'certificats', 'test_services_chain.pem'
+)
```
sebalix, line 18: Could we use [`get_resource_path`](https://github.com/odoo/odoo/blob/12.0/odoo/modules/module.py#L205) for this?

ebill_paynet/data/ir_cron.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sebalix, line 1: Add a copyright in XML files? (not blocking)

ebill_paynet/models/ebill_payment_contract.py
```diff
@@ -0,0 +1,47 @@
...
+        string="Paynet Service",
+        ondelete="restrict",
+    )
+
+    @api.depends('transmit_method_id')
+    def _compute_is_paynet_contract(self):
+        for record in self:
+            record.is_paynet_contract = (
+                record.transmit_method_id
+                == self.env.ref('ebill_paynet.paynet_transmit_method')
```
sebalix: Nitpick: maybe put the call to `ref` outside the `for` loop to get the comparison on one line.

requirements.txt
```diff
@@ -0,0 +1,6 @@
+# For ebill_paynet
+zeep==3.4.0
+# For tests
+freezegun==0.3.12
+vcrpy==2.0.1
+xmlunittest==0.5.0
```
sebalix, line 6: Add `jinja2>=2.10` as it was mentioned in a comment above.

ebill_paynet/models/paynet_invoice_message.py
```diff
@@ -0,0 +1,119 @@
...
+    response = fields.Text('Response recieved')
+    shipment_id = fields.Char(size=24, help='Shipment Id on Paynet service')
+
+    def _get_ic_ref(self):
+        return 'SA%012d' % self.id
+
+    @api.multi
+    def send_to_paynet(self):
+        for message in self:
+            message._generate_payload()
```
sebalix: Nitpick: it's not obvious that some part of the code need this method to be called at first (here and in the tests). I would be explicit by returning the payload and use it in the following lines (we could keep it in the record too to keep a trace of what has been generated). But this implies to put an `self.ensure_one()` in `_generate_payload`.
A matter of taste I suppose :) (not blocking)

ebill_paynet/models/paynet_service.py
```diff
@@ -0,0 +1,223 @@
...
+import os
+
+from lxml import etree
+from odoo import api, fields, models
+from ..components.api import PayNetDWS
+from odoo.exceptions import UserError
+from zeep.exceptions import Fault
+
+SYSTEM_PROD_URL = 'https://dws.paynet.ch/DWS/DWS'
+SYSTEM_TEST_URL = 'https://dws-test.paynet.ch/DWS/DWS'
```
sebalix, line 14: <s>I suggest to put that in `ir.config_parameter`?</s> Doesn't worth it I suppose.

ebill_paynet/models/paynet_service.py
```diff
@@ -0,0 +1,223 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+class PaynetService(models.Model):
+    _name = "paynet.service"
+    _description = "Paynet service configuration"
+
+    name = fields.Char()
+    url = fields.Char(compute='_compute_url', store=True)
```
sebalix, line 27: We already got some trouble with a `store=True` on such configuration fields (which changes when restoring a dump on integration/test/dev). Is it required?

ebill_paynet/models/paynet_service.py
```diff
@@ -0,0 +1,223 @@
...
+        self.ensure_one()
+        dws = PayNetDWS(self.url, self.use_test_service)
+        try:
+            res = dws.client.service.getShipmentContent(
+                Authorization=dws.authorization(self.username, self.password),
+                ShipmentID=shipment_id,
+            )
+        except Fault as e:
+            dws.handle_fault(e)
+            # raise
```
sebalix: Useless line?

ebill_paynet/tests/test_ebill_paynet.py
```diff
@@ -0,0 +1,119 @@
...
+from ..components.api import PayNetDWS
+from odoo.tests.common import SingleTransactionCase
+
+from .common import recorder
+
+
+class TestEbillPaynet(SingleTransactionCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
```
sebalix, line 15: `cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True))`

requirements.txt
TDu: I did not add it because it is a base requirement for Odoo 12.0 https://github.com/odoo/odoo/blob/44123897329c13fc41801f626616c43a24ca13f5/requirements.txt#L12
But added the comment above because internally we are lagging behind...

ebill_paynet/models/paynet_service.py
TDu: I would say not worth it because this standard paynet urls they change them, it changes for all their customer I guess

ebill_paynet/models/paynet_service.py
TDu: It was a reminder for myself :  should it be raised and handled further up ?

ebill_paynet/components/api.py
TDu: Yes we can, changed

ebill_paynet/tests/test_invoice_message.py
TDu: added

ebill_paynet/components/api.py
TDu: Done

ebill_paynet/components/api.py
TDu: Thanks fixed

ebill_paynet/models/ebill_payment_contract.py
TDu: Yes changed

ebill_paynet/models/paynet_invoice_message.py
TDu: Sounds like good taste :wink: so simplifying

ebill_paynet/data/ir_cron.xml
TDu: I don't see any point doing it, except if there is some specific code in there, which is not the case.

yvaucher:
Adding WIP for the TODO list

TDu:
The development of this module has been moving forward on v13 with many improvements.
It would be better to back port from v13, instead of using this.
So closing it.

PR #506, opened 2019-08-28
[12.0] Remove support for 6 digits postal account number

yvaucher:
Postal account numbers with 6 digits only have disappeared.


yvaucher:
included in #509

PR #508, opened 2019-09-02
[MIG] l10n_ch_account_reconcile_isr: Migration to odoo 12

asaunier:
Migration from Odoo 10 https://github.com/OCA/l10n-switzerland/tree/10.0/l10n_ch_account_reconcile_isr


vrenaville:
It need to be adapt , seems that function auto-reconcile do not exist anymore, need to be adapt with account_reconcile_model

grindtildeath:
Changes on https://github.com/OCA/l10n-switzerland/pull/522 and https://github.com/OCA/account-reconcile/pull/292 should make this module obsolete as reconciliation must work out of the box.

PR #509, opened 2019-09-03
[12.0] use core l10n ch fields

yvaucher:
Rework of l10n_ch_base_bank and it's child modules.

- l10n_ch_payment_slip
- l10n_ch_pain_base
- l10n_ch_pain_credit_transfert

Odoo added some `l10n_ch_*` fields to manage ISR. Use them when possible.

* Move dependency on account_payment_order on l10n_ch_pain_base module to make it possible to install l10n_ch_payment_slip without pull all account_payment_order modules.
*  On res.partner.bank `ccp` is replaced by `l10n_ch_postal`
* On res.bank `ccp` field is dropped as it was anyway copied onto `res.partner.bank`
* `reference_type` from `account_payment_order` is not used anymore. ISR reference are detected based on the content of field `reference`
* Drop support of 6 digits postal account numbers
* Add a dedicated field to register ISR customer ID (rework on disambiguation of ISR fields)


l10n_ch_base_bank/models/bank.py
```diff
@@ -1,86 +1,62 @@
...
+def validate_l10n_ch_postal(postal_acc_number):
+    """Check if the string postal_acc_number is a valid postal account number,
+    i.e. it only contains ciphers and is last cipher is the result of a
+    recursive modulo 10 operation ran over the rest of it. Shorten form with -
+    is also accepted.
+    Raise a ValidationError if check fails
+    """
+    if not postal_acc_number:
+        raise ValidationError(_("There is no postal account number."))
+    if re.match('^[0-9]{2}-[0-9]{1,6}-[0-9]$', postal_acc_number or ''):
```
i-vyshnevska: do you need 'or' here?

i-vyshnevska (approved):
LG

l10n_ch_base_bank/models/bank.py
yvaucher: You are right, there is already a check on `not postal_acc_number` thus the or is useless.

yvaucher: Fixed

Martronic-SA:
@yvaucher 

> On res.bank ccp field is dropped as it was anyway copied onto res.partner.bank

l10n_ch_bank was adding the ccp in each res.bank and it now broken with this PR. (see https://github.com/OCA/l10n-switzerland/pull/479#pullrequestreview-285937115).
It allowed to have the ccp on partner bank filled automatically.
Now the user will have to look for the ccp each time he adds a partner bank. Is that ok?

yvaucher:
@Martronic-SA The case when you enter only the name of bank to setup the bank account of your supplier would be very rare.

Usually your supplier gives you at least a postal account number or an IBAN, thus the auto-fill from bank selection is quite useless and add unecessary complexity. Moreover, AFAIK the list of ccp we were loading with l10n_ch_bank module are not reflecting the reality.

PR #510, opened 2019-09-05
[12.0] l10n_ch_payment_slip - Add partner reference in ISR references in order to help reconciliation

yvaucher:
New ISR format is as follow:

```
110011215040203632940468930
\_________/\_____/\______/|
     1        2       3   4

(1) ISR-B Customer ID number (size: may vary)
(2) Partner reference filled by 0 on the left (size: 7)
(3) Invoice reference filled by 0 on the left (size: what is left)
(4) Control digit (size: 1)
```

Sizes truncate the refs from the left

A partner ref such as 10001234567 will become 1234567
Partner ref size is hardcoded to 7 to help search on partner.

This will help reconciliation when customer copies an old reference
of a closed invoice.


l10n_ch_payment_slip/tests/test_payment_slip.py
```diff
@@ -136,6 +136,62 @@ def test_slip_validity(self):
...
+        # partner ref without num
+        self.env.ref('base.res_partner_12').ref = 'alpha'
+        invoice = self.make_invoice()
+        self.assertTrue(invoice.isr_reference)
+        reference = invoice.isr_reference.replace(' ', '')
+        self.assertEqual(len(reference), 27)
+        self.assertEqual(reference[:6], '123456')
+        self.assertEqual(reference[6:13], '0' * 7)
+
+        # contains numerics
```
i-vyshnevska:
```suggestion
        # contains alphabetic
```

i-vyshnevska (approved):

l10n_ch_payment_slip/tests/test_payment_slip.py
yvaucher: Here for me the important part is 123 which will become 0000123.
Thus I meant it contains this but not only.

Will change to alphanumeric then.

yvaucher:
```suggestion
        # contains alphanumerics
```

PR #512, opened 2019-09-10
[10.0] [IMP] match partner in same way as lines for reconcilation

i-vyshnevska:
After isr reconciliation lines reconciled but they didn't have partner_id on them, as isr reconciliation totally relies on transaction_ref 
But partner on statement lines also needed for reports and analytic
This  extension searches for partner in same way as it done for isr reconcile

Technical points
method reconciliation_widget_preprocess is copy pasted from odoo code all changes in preparing sql

### DIFF
--- a/l10n_ch_account_reconcile_isr/models/bank_statement.py
+++ b/l10n_ch_account_reconcile_isr/models/bank_statement.py
@@ -1,0 +1,93 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from odoo import api, models
+
+
+class AccountBankStatement(models.Model):
+
+    _inherit = "account.bank.statement"
+
+    @api.multi
+    def reconciliation_widget_preprocess(self):
+        """ ISR reconciliation relies only on transaction_ref
+        so matches of partner_id also should respect it
+
+        Method is copy of parent the _prepare_query was only added
+        """
+        statements = self
+        sql_query = """SELECT stl.id
+                        FROM account_bank_statement_line stl
+                        WHERE account_id IS NULL AND not exists (
+                            select 1 from account_move m
+                            where m.statement_line_id = stl.id
+                            )
+                            AND company_id = %(company)s
+                """
+        params = {
+            'company': self.env.user.company_id.id,
+        }
+        if statements:
+            sql_query += ' AND stl.statement_id IN %(statement)s'
+            params['statement'] = (tuple(statements.ids),)
+        sql_query += ' ORDER BY stl.id'
+        self.env.cr.execute(sql_query, params)
+        st_lines_left = self.env[
+            'account.bank.statement.line'
+        ].browse([line.get('id') for line in self.env.cr.dictfetchall()])
+
+        # try to assign partner to bank_statement_line
+        stl_to_assign_partner = [
+            stl.id for stl in st_lines_left if not stl.partner_id
+        ]
+        refs = list(set([
+            st.name for st in st_lines_left if not st.partner_id
+        ]))
+        acc_cr = st_lines_left[0].journal_id.default_credit_account_id
+        acc_dt = st_lines_left[0].journal_id.default_debit_account_id
+        if st_lines_left and stl_to_assign_partner and refs\
+           and acc_cr and acc_dt:
+
+            sql_query = self._prepare_query()
+            params.update({
+                'accounts': (acc_cr.id, acc_dt.id),
+                'refs': tuple(refs)
+            })
+
+            if statements:
+                sql_query += 'AND stl.id IN %(statement)s'
+                params['statement'] = tuple(stl_to_assign_partner)
+            self.env.cr.execute(sql_query, params)
+            results = self.env.cr.dictfetchall()
+            st_line = self.env['account.bank.statement.line']
+            for line in results:
+                st_line.browse(
+                    line.get('id')
+                ).write({'partner_id': line.get('partner_id')})
+
+        # return original result to respect all the inheritance
+        # bank statement lines with no partner will be assigned in that call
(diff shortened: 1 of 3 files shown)


l10n_ch_account_reconcile_isr/models/bank_statement.py
```diff
@@ -0,0 +1,92 @@
...
+        return {
+            'st_lines_ids': st_lines_left.ids,
+            'notifications': [],
+            'statement_name':
+                len(statements) == 1 and statements[0].name or False,
+            'num_already_reconciled_lines': 0,
+        }
+
+    def _prepare_query(self):
+        """Replase ref on transaction_ref to match aml"""
```
yvaucher:
```suggestion
        """Replace ref on transaction_ref to match aml"""
```

yvaucher:
@i-vyshnevska do you know if we could have copied the value of transaction_ref into ref field instead? Or is ref field already used by some other data in version 10.0 ?

EDIT: current version would make less meddling in the database.

PS: From version 13.0 we will use ref always.

i-vyshnevska:
> @i-vyshnevska do you know if we could have copied the value of transaction_ref into ref field instead? Or is ref field already used by some other data in version 10.0 ?

From what I see, yes, 
it can solve not only issue introduced in PR, all this addon will be not needed
refactoring will need to be done for few addons, base_transaction_id including

yvaucher:
@i-vyshnevska base_transaction_id will be dropped in version 13.0
But, for version 10.0 we can use your solution.

l10n_ch_account_reconcile_isr/models/bank_statement.py
```diff
@@ -0,0 +1,92 @@
...
+        return {
+            'st_lines_ids': st_lines_left.ids,
+            'notifications': [],
+            'statement_name':
+                len(statements) == 1 and statements[0].name or False,
+            'num_already_reconciled_lines': 0,
+        }
+
+    def _prepare_query(self):
+        """Replace ref on transaction_ref to match aml"""
```
yvaucher:
```suggestion
        """Replace ref by transaction_ref to match aml"""
```

l10n_ch_account_reconcile_isr/models/bank_statement.py
```diff
@@ -0,0 +1,92 @@
...
+                WHERE (aml.company_id = %s
+                    AND aml.partner_id IS NOT NULL)
+                    AND (
+                        (aml.statement_id IS NULL AND aml.account_id IN %s)
+                        OR
+                        (acc.internal_type IN ('payable', 'receivable')
+                         AND aml.reconciled = false)
+                        )
+                    AND aml.transaction_ref IN %s
+                    """
```
yvaucher, line 92: Wouldn't it be better to still search with and return aml.ref if aml.transaction_ref is not set?

i-vyshnevska: I updated  to respect ref as well, return value ref or transaction_ref is not important, doesn't used  anywhere

tschanzt:
@i-vyshnevska does it also work when using the auto esr reconcile button?

i-vyshnevska:
> @i-vyshnevska does it also work when using the auto esr reconcile button?

yes, that preprocessing started by reconciliation wizard

l10n_ch_account_reconcile_isr/models/bank_statement.py
```diff
@@ -0,0 +1,95 @@
...
+                    AND aml.partner_id IS NOT NULL)
+                    AND (
+                        (aml.statement_id IS NULL
+                            AND aml.account_id IN %(accounts)s)
+                        OR
+                        (acc.internal_type IN ('payable', 'receivable')
+                            AND aml.reconciled = false)
+                        )
+                    AND aml.transaction_ref IN %(refs)s
+                    OR  aml.ref IN %(refs)s
```
grindtildeath: Is this condition needed when you want to match on `transaction_ref` ?

grindtildeath (changes requested):
I'm concerned by the lack of a call to super to match on ref. On a multicompany instance with companies in different countries, one might have to match on ref and the other on transaction_ref and it would be broken as inheritance is not respected.
IMO, a context key must be added on the action to use transaction_ref to enable this specific behaviour.

l10n_ch_account_reconcile_isr/models/bank_statement.py
i-vyshnevska: I updated it with call to super, so now it's not broken
As for different ways to reconcile, there is no option to choose, to do so is better to split this  addon on base and other base will add settings, other will register ways to reconcile for now if you install it will apply everywhere

tschanzt (approved):
Tested it. Worked for me.

l10n_ch_account_reconcile_isr/models/bank_statement.py
```diff
@@ -0,0 +1,93 @@
...
+            results = self.env.cr.dictfetchall()
+            st_line = self.env['account.bank.statement.line']
+            for line in results:
+                st_line.browse(
+                    line.get('id')
+                ).write({'partner_id': line.get('partner_id')})
+
+        # return original result to respect all the inheritance
+        # bank statement lines with no partner will be assigned in that call
+        return super(AccountBankStatement, self). \
```
grindtildeath, line 70: Just wondering, shouldn't you update the result with what's done prior to the call to `super`?
e.g.
```
res = super(AccountBankStatement, self).reconciliation_widget_preprocess()
res.update({
    'st_lines_ids': st_lines_left.ids + res.get('st_lines_ids')
})
```

i-vyshnevska: they both return same lines ids

grindtildeath (approved):

yvaucher (approved):

PR #518, opened 2019-10-02
[12.0] Update travis config to get rid of transifex

yvaucher:
We have some error with the job on transifex, which is now obsolete as we got the translation terms from weblate.

Updated from https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml


yvaucher:
Fast forwarding as it should fix travis state.

PR #522, opened 2019-10-15
[12.0] l10n_ch_payment_slip: Fix ref assignation on move lines

grindtildeath:
Payment slip ref on invoice move was overwritten by the
super call to invoice_validate that was done after the
setting of payment slip ref on the move.

Moreover, the dependency to base_transaction_id is dropped
as the reference is now written on account.move.ref field.

Linked to https://github.com/OCA/account-reconcile/pull/292


yvaucher (approved):

yvaucher:
@OCA-git-bot yup you were too slow :)

PR #524, opened 2019-10-25
[12.0] [FIX] l10n_ch_base_bank l10n_ch_postal added twice on partner bank form

Martronic-SA:
No need to add that field l10n_ch_postal here as long as it is already added by l10n_ch.isr_partner_bank_form


yvaucher (approved):
LGTM

travis errors are not related and due to refactor with removal of base_transaction_id ref

PR #527, opened 2019-11-07
[FIX] Use residual instead of debit, useful for partial payment

vrenaville:
* Purpose
We want to be able to resend a bvr with new amount if it's partially paid
* dev
replace the field debit by the amount_residual

### DIFF
--- a/l10n_ch_payment_slip/models/payment_slip.py
+++ b/l10n_ch_payment_slip/models/payment_slip.py
@@ -132,7 +132,7 @@ class PaymentSlip(models.Model):
         :return: total amount of payment slip
         :rtype: float
         """
-        return self.move_line_id.debit
+        return self.move_line_id.amount_residual
 
     @api.depends('move_line_id',
                  'move_line_id.debit',


l10n_ch_payment_slip/models/payment_slip.py
```diff
@@ -132,7 +132,7 @@ def _compute_amount_hook(self):
         :return: total amount of payment slip
         :rtype: float
         """
-        return self.move_line_id.debit
+        return self.move_line_id.amount_residual
```
grindtildeath, line 135: I'm not really sure if this is ok or not...  :thinking: 

On one hand, it makes sense the company wants to get paid only what's owed, but on the other hand there could be use cases where the company wants their invoices to be fully paid and manage overpayments with credit notes. 

Anyway if the company works with the latter case, IMO there shouldn't have been any reconciliations, so I guess it's ok to use `amount_residual` all the time...

grindtildeath (commented):
Please increase module version number :)

We probably also want to forward port this to later versions.

yvaucher:
@grindtildeath no need for module increase version, it can be done on merge by @OCA-git-bot

yvaucher (approved):
LGTM

About @grindtildeath concern, I think that amount_residual cover both cases.

grindtildeath (approved):

PR #534, opened 2019-11-23
13.0 mig l10n ch pain credit transfer

robertrottermann:
work in progress will need dependencies


Martronic-SA (approved):
Tested and worked as expected

PR #535, opened 2019-12-19
[12.0] payment_slip: Fix variable referenced before assignment

grindtildeath:
Fix variable referenced before assignment when currency is not in (EUR, CHF)

### DIFF
--- a/l10n_ch_payment_slip/models/invoice.py
+++ b/l10n_ch_payment_slip/models/invoice.py
@@ -64,7 +64,10 @@ class FutureAccountInvoice(models.Model):
         for record in self:
             isr_subs = False
             isr_subs_formatted = False
-            if record.partner_bank_id:
+            if (
+                record.partner_bank_id and
+                record.currency_id.name in ['EUR', 'CHF']
+            ):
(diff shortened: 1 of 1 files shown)


gurneyalex (approved):

l10n_ch_payment_slip/models/invoice.py
```diff
@@ -70,7 +70,9 @@ def _format_isr_subscription_scanline(isr_subscription):
                     isr_subscription = bank_acc.l10n_ch_isr_subscription_eur
                 elif record.currency_id.name == 'CHF':
                     isr_subscription = bank_acc.l10n_ch_isr_subscription_chf
-                # else we don't format if in another currency as EUR or CHF
+                else:
+                    # we don't format if in another currency as EUR or CHF
+                    isr_subscription = False
```
yvaucher: Better to check record.currency_id.name in ['EUR', 'CHF'] at line 67 ?

grindtildeath:
@yvaucher oca bot didn't care that I push forced the branch with your suggestion, IDK if we still need to change this ? BTW it also seems the `api.depends` is not correct...

PR #536, opened 2020-01-15
[11.0] Migrate l10n_ch_pain_direct_debit and merge with l10n_ch_lsv_dd

ecino:
This allows to generate swiss pain files and simplifies the repository structure, by merging modules `lsv_dd` with `pain_direct_debit` module.


ecino:
@yvaucher Can you have a look at this one?

l10n_ch_pain_direct_debit/__manifest__.py
```diff
@@ -1,17 +1,29 @@
-# -*- coding: utf-8 -*-
 # © 2016 Akretion (Alexis de Lattre <alexis.delattre@akretion.com>)
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-
+# pylint: disable=C8101
```
yvaucher, line 3: Weird, C8101 is a check about presence of Odoo Community Association (OCA) in the author list.

My guess is that by removing the space between `Association` and `(OCA)` you get a lint error. 

This is the reference:

https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py

l10n_ch_pain_direct_debit/models/account_payment_order.py
```diff
@@ -0,0 +1,452 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
yvaucher, line 21: You might want to change this to the shord format for headers.

yvaucher (changes requested):
Code review only,

LGTM only a couple of module "cosmetics' otherwise good to be merged.

ecino:
Thanks, I will make the changes. By testing we saw that the format for Swiss banks direct debit has some specificities that we have to change here and there, so we should wait that we fix these before we merge. I will tell you when it's done and properly tested.

PR #538, opened 2020-01-23
[12.0] [FIX] l10n_ch_payment_slip correct view dependency

Martronic-SA:
The view hides then button "isr_print" and then should inherit on the view that puts that button, which is l10n_ch.isr_invoice_form
That's fine as long as the module l10n_ch_payment_slip depends on l10n_ch


yvaucher (approved):
It makes sense with the new dependency.

PR #539, opened 2020-02-04
[12.0] [FIX] l10n_ch_payment_slip : Fix javascript

Martronic-SA:
Clicking on "Print ISR" from an invoice will throw the following error (visible in the developer console): "The ActionManager can't handle reports of type reportlab-pdf"
It seems that the javascrpt had not been tested and not ported to v12.0 framework.
Here is the PR for v12.0


yvaucher (approved):
Code LGTM

tested on runbot

The previous version was failing silently when clicking on the Print button.
This fixes it.


Thanks @Martronic-SA

PR #540, opened 2020-02-07
[12.0] account_invoice_with_payment - Do not forward default_type context key to the report

yvaucher:
The context key 'default_type' can be added by an action.
It is the case if you create an invoice after having navigated
through the "Invoiced" smart button on partner form.
This context key is only intended for the invoice and shall not be
passed to the report.

### DIFF
--- a/l10n_ch_invoice_with_payment/models/report.py
+++ b/l10n_ch_invoice_with_payment/models/report.py
@@ -9,12 +9,17 @@ class IrActionsReport(models.Model):
 
     @api.multi
     def render_qweb_pdf(self, res_ids=None, data=None):
+        ctx = self.env.context
+        # Remove context key that comes from invoice action
+        if ctx.get('default_type'):
+            ctx = ctx.copy()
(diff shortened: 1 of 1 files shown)


sebalix (approved):

l10n_ch_invoice_with_payment/models/report.py
```diff
@@ -9,12 +9,17 @@ class IrActionsReport(models.Model):
 
     @api.multi
     def render_qweb_pdf(self, res_ids=None, data=None):
+        ctx = self.env.context
+        # Remove context key that comes from invoice action
+        if ctx.get('default_type'):
+            ctx = ctx.copy()
+            ctx.pop('default_type')
```
i-vyshnevska, line 16: maybe this should go after line 19?

yvaucher: This context key comes from `account.invoice` the current object is already an `ir.actions.report` to me it shouldn't be there at all.

PR #516, opened 2019-10-01
[13.0][MIG] l10n_ch_base_bank

i-vyshnevska:
migrate l10n_ch_base_bank
~it blocked by https://github.com/odoo/odoo/pull/37651~
it blocked by  https://github.com/odoo/odoo/pull/49431

PR #541, opened 2020-02-16
[13.0][MIG] l10n_ch_pain_base

pegonzalezsredcor:
Module installable in v13
Adapted account.move.line.move_id
Adapted _is_isr_ref instead of is_isr_reference
Waiting on the following to do a final functional revision:
- [x] ~~#516~~ #550
- [x] https://github.com/OCA/bank-payment/pull/651 

EDIT: @yvaucher  list dependencies


yvaucher:
restarting travis jobs now that bank-payment dependency is merged

pedrobaeza:
@yvaucher what is the status of this? These modules are installed in the OCA instance and we need it for the DB migration

yvaucher:
@pedrobaeza did not have the chance to dive into this module specificaly, we have been using the `account_sepa` enterprise modules lately. So I would be ok to do some optimistic merge on this. However, AFAIK those modules have worked for standard case but they might not take into account the last changes in `l10n_ch`. Which is currently still a bit messy. And this state of the official module has impact on most modules in this localization.

pedrobaeza:
Well, we can put these PRs in the OCA instance, but I don't feel comfortable enough with the result for this part, but this is out of the scope of the bid for migrating the DB, as they were added later and they are country specific. It would be good that you check them when possible.

yvaucher (approved):
LGTM

local instrument is correctly set to CH01 when we have an ISR reference.
And in such case the content for the acc number is taken from `l10n_ch_postal`

Which mean it will work for most simple cases.

Plus I checked that local instrument is set at payment line level, which will allow more complexes cases of multi-payments with different types in a single file.

l10n_ch_pain_base/models/account_move_line.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import models
+
+
+class AccountMoveLine(models.Model):
+    _inherit = "account.move.line"
+
+    def _prepare_payment_line_vals(self, payment_order):
+        vals = super()._prepare_payment_line_vals(payment_order)
+        if self.move_id and self.move_id._is_isr_ref():
```
yvaucher: `is_isr_ref` is not enough here. It only checks that we have a 27 digits reference.

What you are looking for is to know if the tuple (bank_account, ref) is of ISR type.

The bank account must be an ISR issuer. Postal account number format starting by `01` or `03`
https://github.com/odoo/odoo/blob/13.0/addons/l10n_ch/models/res_bank.py#L52

Thus you want something like this: (probably needs checks that self.move_id and self.move_id.partner_bank_id are checked.)

```suggestion
        if self.move_id.partner_bank_id._is_isr_issuer() and self.move_id._is_isr_ref():
```

yvaucher (changes requested):
Just spotted one issue while reviewing the last changes.

With the arrival of QRR, you can have 27 digits refs but without using an ISR subscription number (01-162-8 format) you will use a QR-IBAN instead.

To me that fix can be applied in an other PR nevertheless.

yvaucher (approved):

l10n_ch_pain_base/models/account_move_line.py
pegonzalezs: @yvaucher ... doing a functional revision I get next error:

image

so a new commit with the fix will be included

robertrottermann (approved):
LGTM

i-vyshnevska:
@pegonzalezsredcor can you rebase PR, dependent PR's had been merged

Martronic-SA (approved):
Tested and worked as expected

Martronic-SA (changes requested):
https://github.com/OCA/l10n-switzerland/pull/576 has to be ported here also

yvaucher:
@pegonzalezs can you rebase to make sure the CI passes now that all dependencies are merged?

PR #544, opened 2020-03-26
[IMP] add possibility to send file without BIC set

vrenaville:
* Purpose
WE need to be able to send SEPA file without BIC, it's not necessary to make a payment.
I have add it as an option, by default is checked (to keep the same process), but can be deactivated

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -68,7 +68,8 @@ class AccountPaymentOrder(models.Model):
             if bank_line.local_instrument == 'CH01':
                 # Don't set the creditor agent on BVR/CH01 payments
                 return True
-            elif not partner_bank.bank_bic:
+            elif not partner_bank.bank_bic and \
+                    self.payment_method_id.bic_required:
                 raise UserError(_(
                     "For pain.001.001.03.ch.02, for non-BVR payments, "
                     "the BIC is required on the bank '%s' related to the "
--- a/l10n_ch_pain_base/__openerp__.py
+++ b/l10n_ch_pain_base/__openerp__.py
@@ -17,6 +17,7 @@
     "data": [
         'views/account_payment_line.xml',
         'views/bank_payment_line.xml',
+        'views/account_payment_method_view.xml'
         ],
     'installable': True,
 }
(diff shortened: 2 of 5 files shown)


l10n_ch_pain_base/models/account_payment_order.py
```diff
@@ -68,7 +68,8 @@ def generate_party_agent(
             if bank_line.local_instrument == 'CH01':
                 # Don't set the creditor agent on BVR/CH01 payments
                 return True
-            elif not partner_bank.bank_bic:
+            elif not partner_bank.bank_bic and \
+                    self.payment_method_id.bic_required:
```
sebalix, line 72: As we are in a `@api.model` the reference to `self.payment_method_id` is not supposed to work here.

sebalix: Answer: It is working as the caller `generate_party_block()` is decorated with `@api.multi` so `self` represents a recordset, but it's not expected as input.
I'll fix this in `bank-payment` modules while porting this code to 12.0, I don't think it worth to pass time to "fix it" on 9.0.

yvaucher, line 72: The rule must be about using an IBAN or QR-IBAN.

If IBAN -> we don't need a BIC (it can be deduced from IBAN)

```suggestion
            elif not partner_bank.bank_bic and not partner_bank.type == 'iban':
```

yvaucher (changes requested):
@vrenaville 

TL;DR

The check on the BIC should be improved. We don't need an additional field.

----

Reading the Development Guidelines, it seems the BIC can always be omitted since a change since 2019 for payment Type 3 (the main use of this module).

https://www.six-group.com/dam/download/banking-services/interbank-clearing/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf

> 22.09.2017
Section 2.2.1 and 2.3.3: With payment type 3 the Creditor Agent information (IID/BIC)can be omitted with effect from **01.01.2019**, where the IBAN/QR-IBAN is sent. 

Here I don't know if your change is to alter the Debtor Agent or the Creditor Agent or both.


For creditor agent:

> BIC: 
Type 3(V2), 4(V1): May be present. When thepayment is executed, the Creditor Agent isalways worked out from the IBAN (CH/LI), if available.
Type 5: May be present. With payment type 5, the Creditor Agent is always worked out from the IBAN for execution of the payment.
Type 6: Must contain "BIC International".

PR #546, opened 2020-04-23
[12.0][MIG] Migrate l10n_ch_import_pain002

philhug:


philhug:
missing dependency `account_payment_line_cancel`: hold for now

ecino:
Hello @philhug, thanks for your request. I want to point out that we are currently migrating all FDS and related modules for v12 and it required some big refactors. We are in the process of validating the changes before we open the pull requests (there will be several).

Here is our take on the import_pain002 migration: https://github.com/CompassionCH/l10n-switzerland/tree/12.0-payment-return-sepa

We renamed the module payment_return_sepa and removed the dependency on the account_payment_line_cancel module. It now takes advantage of the `account_payment_return_import_iso20022` available on OCA/account-payment

PR #548, opened 2020-04-23
[12.0][MIG] Migrate l10n_ch_fds_upload_sepa

philhug:
Depends on #546


ecino:
We have some more changes for v12 coming for FDS modules ready here: https://github.com/CompassionCH/l10n-switzerland/tree/12.0-fds
We just need to test more before we were opening pull requests.

PR #555, opened 2020-07-24
[FIX] l10n_ch_qr_bill: duplicated bank field

yvaucher:
partner_bank_id field is already present on the form but invisible. Change the visibility attribute instead of re-creating it


thomich:
:+1:

PR #556, opened 2020-07-28
[ADD] l10n_ch_qr_bill_transaction_id - glue module

yvaucher:
New module to use transaction_id with QRR process for reconciliation

### DIFF
--- a/l10n_ch_qr_bill_transaction_id/models/account_invoice.py
+++ b/l10n_ch_qr_bill_transaction_id/models/account_invoice.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+# Copyright 2020 Camptocamp SA
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, models
+
+
+class AccountInvoice(models.Model):
+    _inherit = 'account.invoice'
+
+    @api.multi
+    def _propagate_qrr(self):
+        """Prepare the propagation of QRR by copying it to transaction_id
+
+
+        For suppliers invoices: the QRR reference is stored in `name`
+        field of the invoice.
+
+        For customers invoices: the QRR will be stored in `l10n_ch_qrr`
+        but it will only be filled once the move is generated.
+
+        """
+        self.ensure_one()
+        if self.partner_bank_id._is_qr_iban():
+            if self.type in ('in_invoice', 'in_refund'):
+                if self._is_qrr(self.reference):
+                    self.transaction_id = self.reference.replace(" ", "")
+            else:
+                # Generate a placeholder as the QRR is not generated yet
+                self.transaction_id = "_l10n_ch_qrr_{}".format(self.id)
+
+    @api.multi
+    def _post_propagate_qrr(self):
+        """Replace the placeholders by the QRR"""
+        MoveLine = self.env["account.move.line"]
+        for inv in self:
+            placeholder = "_l10n_ch_qrr_{}".format(inv.id)
+            if inv.transaction_id and inv.transaction_id == placeholder:
+                inv.transaction_id = inv.l10n_ch_qrr
+                lines = MoveLine.search(
+                    [("transaction_ref", "=", placeholder)]
+                )
+                lines.write({"transaction_ref": inv.l10n_ch_qrr})
+
+    @api.multi
+    def action_move_create(self):
+        """Prepare the copy of the QRR in to transaction_id
+
+        Which will be later transfered in `finalize_invoice_move_lines`
+        to `transaction_ref` on `account.move.line`
+
+        """
+        for inv in self:
+            inv._propagate_qrr()
+        res = super(AccountInvoice, self).action_move_create()
+        self._post_propagate_qrr()
+        return res
--- a/l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
+++ b/l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
@@ -33,7 +33,7 @@ class TestTransactionID(AccountingTestCase):
         )
         self.invoice1 = self.create_invoice('base.CHF')
 
-    def create_invoice(self, currency_to_use='base.CHF'):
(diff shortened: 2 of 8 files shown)


l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
```diff
@@ -0,0 +1,104 @@
...
+                "street": "Route de Berne 88",
+                "street2": "",
+                "zip": "2000",
+                "city": "Neuchâtel",
+                "country_id": self.env.ref('base.ch').id,
+            }
+        )
+        self.invoice1 = self.create_invoice('base.CHF')
+
+    def create_invoice(self, currency_to_use='base.CHF'):
```
grindtildeath:
```suggestion
    def create_invoice(self, currency_xmlid='base.CHF'):
```
or require to pass the recordset directly?

l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
```diff
@@ -0,0 +1,104 @@
...
+            [('user_type_id', '=', acc_type.id)], limit=1
+        )
+        invoice = (
+            self.env['account.invoice']
+            .with_context(default_type='out_invoice')
+            .create(
+                {
+                    'type': 'out_invoice',
+                    'partner_id': self.customer.id,
+                    'currency_id': self.env.ref(currency_to_use).id,
```
grindtildeath:
```suggestion
                    'currency_id': self.env.ref(currency_xmlid).id,
```

l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
```diff
@@ -0,0 +1,104 @@
...
+    def test_swissQR_qriban(self):
+        # Now use a proper QR-IBAN, we are good to print a QR Bill
+        # In this case the transaction_id must be set
+        qriban_account = self.create_account(QR_IBAN)
+        self.assertTrue(qriban_account.acc_type, 'qr-iban')
+        self.invoice1.partner_bank_id = qriban_account
+        self.invoice1.action_invoice_open()
+        self.assertEqual(
+            self.invoice1.transaction_id, self.invoice1.l10n_ch_qrr
+        )
```
grindtildeath, line 104: Would be great to assert the values written on `account.move` + `account.move.line`

l10n_ch_qr_bill_transaction_id/models/account_invoice.py
```diff
@@ -0,0 +1,57 @@
...
+
+        """
+        self.ensure_one()
+        if self.partner_bank_id._is_qr_iban():
+            if self.type in ('in_invoice', 'in_refund'):
+                if self._is_qrr(self.reference):
+                    self.transaction_id = self.reference.replace(" ", "")
+            else:
+                # Generate a placeholder as the QRR is not generated yet
+                self.transaction_id = "_l10n_ch_qrr_{}".format(self.id)
```
grindtildeath, line 30: Could be wrapped in a small function so you can call it as well on line 37 :wink:

l10n_ch_qr_bill_transaction_id/models/account_invoice.py
```diff
@@ -0,0 +1,57 @@
...
+    @api.multi
+    def _propagate_qrr(self):
+        """Prepare the propagation of QRR by copying it to transaction_id
+
+
+        For suppliers invoices: the QRR reference is stored in `name`
+        field of the invoice.
+
+        For customers invoices: the QRR will be stored in `l10n_ch_qrr`
+        but it will only be filled once the move is generated.
```
grindtildeath, line 20: This comment is a bit confusing as `name` isn't used in the function below for supplier invoices.
And AFAIU this will not only "prepare" the propagation for in_* types but write the final value, isn't it?

l10n_ch_qr_bill_transaction_id/models/account_invoice.py
```diff
@@ -0,0 +1,57 @@
...
+                # Generate a placeholder as the QRR is not generated yet
+                self.transaction_id = "_l10n_ch_qrr_{}".format(self.id)
+
+    @api.multi
+    def _post_propagate_qrr(self):
+        """Replace the placeholders by the QRR"""
+        MoveLine = self.env["account.move.line"]
+        for inv in self:
+            placeholder = "_l10n_ch_qrr_{}".format(inv.id)
+            if inv.transaction_id and inv.transaction_id == placeholder:
```
grindtildeath, line 38: As I understand this condition will only be verified for customers invoices/refunds, because transaction_id will already include the QR ref for suppliers?
Maybe state the function does it only for out_* types or use a `filtered` ?

l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
```diff
@@ -0,0 +1,104 @@
...
+        self.env.user.company_id.partner_id.write(
+            {
+                "street": "Route de Berne 88",
+                "street2": "",
+                "zip": "2000",
+                "city": "Neuchâtel",
+                "country_id": self.env.ref('base.ch').id,
+            }
+        )
+        self.invoice1 = self.create_invoice('base.CHF')
```
grindtildeath, line 34: Would be great to add a test for supplier invoice as well :wink:

grindtildeath (commented):
Some nitpicking and questions, but looks good overall. 

Wondering if a glue module is really required though, but you call :wink:

l10n_ch_qr_bill_transaction_id/tests/test_transaction_id.py
yvaucher: recordset wouldn't help getting a default value.

thomich:
hey

that would also be nice, if it will be merged, as all tests passed.

thanks
thomi

yvaucher:
@grindtildeath I think @thomich is right, I prefer to publish this now in the current state. Didn't had much time to do much more on this.

PR #557, opened 2020-07-28
[FIX] add file extension for QR-Bill report

yvaucher:
Be nice with software that rely on file extension to discover the file type.


yvaucher:
fastforwarding for triviality and the module is brand new

PR #558, opened 2020-08-04
[12.0][IMP] add possibility to send file without BIC set

sebalix:
Forward port of https://github.com/OCA/l10n-switzerland/pull/544

Depends on https://github.com/OCA/bank-payment/pull/724 (to merge first) to get `api.multi` decorated method instead of `api.model`.

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -58,7 +58,6 @@ class AccountPaymentOrder(models.Model):
             gen_args,
         )
 
-    @api.model
     def generate_party_agent(
             self, parent_node, party_type, order, partner_bank, gen_args,
             bank_line=None):
@@ -68,7 +67,8 @@ class AccountPaymentOrder(models.Model):
             if bank_line.local_instrument == 'CH01':
                 # Don't set the creditor agent on ISR/CH01 payments
                 return True
-            elif not partner_bank.bank_bic:
+            elif not partner_bank.bank_bic and \
+                    any(self.mapped("payment_method_id.bic_required")):
                 raise UserError(_(
                     "For pain.001.001.03.ch.02, for non-ISR payments, "
                     "the BIC is required on the bank '%s' related to the "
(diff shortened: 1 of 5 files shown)


sebalix:
ping @vrenaville

l10n_ch_pain_base/models/account_payment_order.py
```diff
@@ -68,7 +67,8 @@ def generate_party_agent(
             if bank_line.local_instrument == 'CH01':
                 # Don't set the creditor agent on ISR/CH01 payments
                 return True
-            elif not partner_bank.bank_bic:
+            elif not partner_bank.bank_bic and \
+                    any(self.mapped("payment_method_id.bic_required")):
```
yvaucher, line 71: The rule must be about using an IBAN or QR-IBAN.

If IBAN -> we don't need a BIC (it can be deduced from IBAN)

```suggestion
            elif not partner_bank.bank_bic and not partner_bank.type == 'iban':
```

yvaucher (changes requested):
@vrenaville 

TL;DR

The check on the BIC should be improved. We don't need an additional field.

----

Reading the Development Guidelines, it seems the BIC can always be omitted since a change since 2019 for payment Type 3 (the main use of this module).

https://www.six-group.com/dam/download/banking-services/interbank-clearing/en/standardization/iso/swiss-recommendations/implementation-guidelines-ct.pdf

> 22.09.2017
Section 2.2.1 and 2.3.3: With payment type 3 the Creditor Agent information (IID/BIC)can be omitted with effect from **01.01.2019**, where the IBAN/QR-IBAN is sent. 

Here I don't know if your change is to alter the Debtor Agent or the Creditor Agent or both.


For creditor agent:

> BIC: 
Type 3(V2), 4(V1): May be present. When thepayment is executed, the Creditor Agent isalways worked out from the IBAN (CH/LI), if available.
Type 5: May be present. With payment type 5, the Creditor Agent is always worked out from the IBAN for execution of the payment.
Type 6: Must contain "BIC International".

l10n_ch_pain_base/models/account_payment_order.py
yvaucher, line 71: No need for an additional field

```suggestion
            elif not partner_bank.bank_bic and not partner_bank.type == "iban":
```

yvaucher (changes requested):
~deleted~ see upper

PR #562, opened 2020-09-16
[8.0][MIG] l10n_ch_qr_bill

SimoneVagile:
Backporting from 10.0 version to 8.0.


l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -20,12 +26,10 @@ def _is_qr_iban(self):
             return False
 
         self.ensure_one()
-
-        iid_start_index = 4
-        iid_end_index = 8
-        iid = self.sanitized_acc_number[iid_start_index : iid_end_index + 1]
+        sanitized_acc_number = self._sanitize_account_number(self.acc_number)
+        iid = sanitized_acc_number[4:9]
```
fredzamoabg, line 29: _sanitize_account_number method you defined can return false. probably you should check before extracting information like sanitized_acc_number[4:9]

SimoneVagile: [https://github.com/odoo/odoo/blob/cca4fb071509164493d0febabf53eb43168f6fb6/odoo/addons/base/res/res_bank.py#L61](https://github.com/odoo/odoo/blob/cca4fb071509164493d0febabf53eb43168f6fb6/odoo/addons/base/res/res_bank.py#L61) res_bank.acc_number is a required field and in this case is always enhanced.
Indeed, this is a copy of the method developed in res.bank model on v10 [https://github.com/odoo/odoo/blob/cca4fb071509164493d0febabf53eb43168f6fb6/odoo/addons/base/res/res_bank.py#L9](https://github.com/odoo/odoo/blob/cca4fb071509164493d0febabf53eb43168f6fb6/odoo/addons/base/res/res_bank.py#L9).

The line raises exception correctly if `sanitized_acc_number` is False (that means there's something wrong).

fredzamoabg: ok

fredzamoabg (approved):

yvaucher (approved):

yvaucher:
@SimoneVagile could you rebase? travis unit tests should be fixed now for bank statement import. Will still fail on linting but it's less critical.

yvaucher:
Ok, looking at travis CI state on branch 8.0, I don't think it's worth spending too much time on it.

I'll merge this to make the module available. I'll check if I can post fix a couple of things.

PR #563, opened 2020-09-17
[10.0][BACKPORT] l10n_ch_qriban and include it in l10n_ch_qr_bill

yvaucher:
On version 13.0 following on support ticket Odoo added the module `l10n_ch_qriban`

This adds a field on `res.partner.bank` to store an QR-IBAN beside an IBAN.
QR-IBAN can be considered as aliases for IBAN with the sole purpose of being used with QRR.

This change in data model allows to have both information in order to do the reconciliation later using a single journal.

### DIFF
--- a/l10n_ch_qr_bill/__manifest__.py
+++ b/l10n_ch_qr_bill/__manifest__.py
@@ -15,6 +15,7 @@
     'data': [
         'data/paperformat.xml',
         'views/account_invoice_view.xml',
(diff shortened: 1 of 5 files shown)


thomich:
hello

whats about this PR, would be nice if it will pass the tests and so we have it for o10.

thx
thomi

yvaucher:
@thomich it just needs some reviews

PR #566, opened 2020-10-26
[MIG][12.0] l10n_ch_pain_direct_debit

ecino:

PR #565, opened 2020-10-26
[MIG][12.0] l10n_ch_bank_statement_import_postfinance

ecino:

PR #545, opened 2020-04-23
[12.0][MIG] Migrate l10n_ch_fds_postfinance

philhug:

PR #570, opened 2020-10-26
[12.0] Migrate all FDS modules

ecino:
This migrates all FDS modules previously available and adds a few new ones that takes advantage of payment returns in order to import payment returns from FDS.

Depends on #565 #569 and #566 

This would override #545


l10n_ch_fds_postfinance/wizards/fds_files_import_from_fds_wizard.py
```diff
@@ -113,11 +113,11 @@ def import_button(self):
                 self.process_files(file)
 
             self.state = 'done'
-        except Exception as e:
+        except:
```
yvaucher, line 117: Are we sure we want to catch all exceptions ? isn't there a subset of exception that are ok during import?

l10n_ch_fds_postfinance/wizards/fds_files_import_from_fds_wizard.py
```diff
@@ -110,6 +110,8 @@ def import_button(self):
             # process the files (done by the childrens of this module)
             for file in fds_files_ids:
                 self.process_files(file)
+                # We save the file if correctly processed
+                self.env.cr.commit()
```
yvaucher, line 114: It's usually a bad idea to commit the main cursor.

I see we have the same at line https://github.com/OCA/l10n-switzerland/pull/570/commits/ad7fc00914723b2aaf24cd586253cbbb2aa10e07#diff-8fcb44095362db5b558733ac48f92b9353dba18438173751467d5239e3e41d17L72

Maybe you are looking at the savepoint case.
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#never-commit-the-transaction

yvaucher (changes requested):
I'm only concerned about the `cr.commit` which is risky.

You either want a new cursor or you want a savepoint.

l10n_ch_fds_postfinance/wizards/fds_files_import_from_fds_wizard.py
yvaucher: @ecino gentle ping about this last issue on this PR

PR #574, opened 2020-11-05
[12.0] [ADD] l10n_ch_fix_isr_reference

yvaucher:
Fix computation of l10n_ch_isr_number which is broken in Odoo core 12.0


yvaucher:
Cancelled OCA-git-bot build to make a rebase first.

PR #576, opened 2020-11-25
12.0 add QRR pain001

davidwul:
Add the ability to include a QR invoice with QRR on a payement order and produce a pain001 expected (according to SIX implementation-guidelines-ct)

PR #579, opened 2020-12-04
[11.0][Backport] add QRR pain001

leemannd:
Before merging this, we need to check if the fix from odoo is working without this -> https://github.com/odoo/enterprise/pull/14924/files


yvaucher:
@leemannd enterprise module and this OCA module are 2 separate flows for the same thing.

Otherwise, as this fix is obviously based on #576 you should author it properly using `git commit --author='...'` or by doing a `git cherry-pick`

leemannd:
Closing as there are no news on this one

PR #580, opened 2020-12-07
[12.0] l10n_ch_payment_slip hide core "ISR Print" button

Martronic-SA:
"ISR Print" core button is not hidden properly because invisible attribute is set from attrs="{invisible..." so we have to remove attrs attribute to make it really invisible.


yvaucher (approved):
:+1:

Just restarted travis which failed for not specific reason.

PR #581, opened 2020-12-10
[13.0] Add l10n_ch_invoice_reports module

ajaniszewska-dev:
aim of this PR is to provide extended invoice reports (invoice with isr and/or qr) and possibility to download it from the portal. Since extended report is merged in memory, it has no xml template assigned to it - therefore it is not possible to render this document in document preview on portal side - we made a decision to keep showing the standard invoice there.


l10n_ch_invoice_reports/models/report.py
```diff
@@ -0,0 +1,72 @@
...
+        qr_pdf,_ = qr_report.render_qweb_pdf(res_ids,data)
+        qr_pdf_io = io.BytesIO(qr_pdf)
+
+        pdf = False
+        if self.report_name == reports[0]:
+            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io])
+        elif self.report_name == reports[1]:
+            pdf = self.merge_pdf_in_memory([invoice_pdf_io, qr_pdf_io])
+        else:
+            pdf =self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io, qr_pdf_io])
```
yvaucher: lint

```suggestion
            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io, qr_pdf_io])
```

l10n_ch_invoice_reports/data/data.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
simahawk: rename this file `reports.xml`

l10n_ch_invoice_reports/models/report.py
```diff
@@ -0,0 +1,72 @@
...
+        invoice_pdf, _ = inv_report.render_qweb_pdf(res_ids, data)
+        invoice_pdf_io = io.BytesIO(invoice_pdf)
+
+        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
+        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
+        isr_pdf_io = io.BytesIO(isr_pdf)
+
+        qr_report = self._get_report_from_name("l10n_ch.qr_report_main")
+        qr_pdf, _ = qr_report.render_qweb_pdf(res_ids, data)
+        qr_pdf_io = io.BytesIO(qr_pdf)
```
simahawk, line 58: no need to close these objects?

l10n_ch_invoice_reports/data/data.xml
```diff
@@ -0,0 +1,39 @@
...
+        report_type="qweb-pdf"
+        string="Invoice with QR"
+        name="l10n_ch_invoice_reports.invoice_qr_report_main"
+        file="l10n_ch_invoice_reports.invoice_qr_report_main"
+        attachment_use="False"
+        attachment=""
+        menu="True"
+        print_report_name="'Invoice-QR-%s' % object.name"
+    />
+    <report
```
simahawk: FTR the `report` tag is deprecated on v14 https://github.com/odoo/odoo/commit/6835aeb0de6895f7f4d6b23e0b4654465ef21d6a

l10n_ch_invoice_reports/models/report.py
```diff
@@ -0,0 +1,75 @@
...
+        reports = [
+            "l10n_ch_invoice_reports.invoice_isr_report_main",
+            "l10n_ch_invoice_reports.invoice_qr_report_main",
+            "l10n_ch_invoice_reports.invoice_qr_isr_report_main",
+        ]
+        ctx = self.env.context
+        # Remove context key that comes from invoice action
+        if ctx.get("default_type"):
+            ctx = ctx.copy()
+            ctx.pop("default_type")
```
TDu: This modified context `ctx` is not used later on!
Unnecessary code ?

l10n_ch_invoice_reports/models/report.py
TDu: They are closed a few lines down https://github.com/OCA/l10n-switzerland/pull/581/files#diff-95e065a95f81606f9d2f3111e3db2866670a00966c9e8442d6168e756046a06eR67

TDu (approved):

l10n_ch_invoice_reports/controllers/portal.py
```diff
@@ -0,0 +1,47 @@
...
+from odoo.http import request
+
+from odoo.addons.account.controllers.portal import PortalAccount
+
+
+class PortalAccount(PortalAccount):
+    @http.route(
+        ["/my/invoices/<int:invoice_id>"], type="http", auth="public", website=True
+    )
+    def portal_my_invoice_detail(
```
TDu, line 10: Looks like you are redefining an existing route from the `account` module.
Hard to tell what changes but it seems to be the pdf download, only.
In that case it would be better/cleaner to not copy paste the existing implementation, but implement the download of the pdf here and use the parent implementation with `super` for the rest

simahawk: I agree w/ @TDu.
Also, when overriding a route method, don't re-define all params. Just use `@http.route()`

simahawk (changes requested):

l10n_ch_invoice_reports/controllers/portal.py
ajaniszewska-dev: here i am changing only parameter (report_ref) with which method is called depending on the type of document. It is not defined anywhere before. Not sure how super() can help. Already talked about that with @TDu on Friday.

http.route fixed.

simahawk (approved):

PR #582, opened 2021-01-13
[10.0] Fix l10n ch QR-IBAN validation

santostelmo:
Only IBANs starting with CH and LI can be considered QR IBANs

### DIFF
--- a/l10n_ch_qr_bill/models/res_bank.py
+++ b/l10n_ch_qr_bill/models/res_bank.py
@@ -24,6 +24,12 @@ class ResPartnerBank(models.Model):
         )
     )
 
+    @api.constrains('l10n_ch_qr_iban')
+    def _check_ch_li_l10n_ch_qr_iban(self):
+        """Validate QR-IBAN"""
+        for record in self:
+            self._validate_qr_iban(record.l10n_ch_qr_iban)
+
     def _check_qr_iban_range(self, iban):
         if not iban or len(iban) < 9:
             return False
@@ -35,6 +41,9 @@ class ResPartnerBank(models.Model):
                 and 30000 <= int(iid) <= 31999)
 
     def _validate_qr_iban(self, qr_iban):
+        if qr_iban and not qr_iban.startswith(('CH', 'LI')):
+            raise ValidationError(_(
+                "Not a valid Switzerland or Liechtenstein QR-IBAN."))
         # Check first if it's a valid IBAN.
         validate_iban(qr_iban)
         # We sanitize first so that _check_qr_iban_range()
@@ -69,6 +78,9 @@ class ResPartnerBank(models.Model):
         """
         return (
             self.acc_type == "iban"
-            and self._check_qr_iban_range(self.sanitized_acc_number)
-            or self.l10n_ch_qr_iban
+            and self.l10n_ch_qr_iban
+            or (
+                self._check_qr_iban_range(self.sanitized_acc_number) and
+                self.sanitized_acc_number.startswith(('CH', 'LI'))
+            )
         )


l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -24,6 +24,18 @@ class ResPartnerBank(models.Model):
         )
     )
 
+    @api.multi
+    @api.constrains('l10n_ch_qr_iban')
```
carlosserratoro, line 27: You constrain depending on ```l10n_ch_qr_iban``` but you are working in the code only with ```l10n_ch_qrr```. Does ```l10n_ch_qr_iban``` depend on the value set on ```l10n_ch_qrr```?

santostelmo: Thanks carlos, this was a mistake. Changing. it should be `l10n_ch_qr_iban` used in the method.

carlosserratoro (approved):

l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -24,6 +24,18 @@ class ResPartnerBank(models.Model):
...
+    def _check_ch_li_l10n_ch_qr_iban(self):
+        """Check if field is populated with Swiss / Liechtenstein qr iban"""
```
yvaucher:
```suggestion
        """Check if field is populated with Swiss / Liechtenstein QR-IBAN"""
```

l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -24,6 +24,18 @@ class ResPartnerBank(models.Model):
         )
     )
 
+    @api.multi
```
yvaucher: `api.constraint` are `api multi` by default. No need to specify it.

l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -69,6 +81,11 @@ def _is_qr_iban(self):
         """
         return (
             self.acc_type == "iban"
-            and self._check_qr_iban_range(self.sanitized_acc_number)
-            or self.l10n_ch_qr_iban
+            and (
+                self.l10n_ch_qr_iban and
+                self.l10n_ch_qr_iban.startswith(('CH', 'LI'))
+            ) or (
```
yvaucher: This should already be validated by the constraint.

```suggestion
            and self.l10n_ch_qr_iban
            or (

File: l10n_ch_qr_bill/models/res_bank.py [hunk 5]
```diff
@@ -24,6 +24,18 @@ class ResPartnerBank(models.Model):
...
+    @api.constrains('l10n_ch_qr_iban')
+    def _check_ch_li_l10n_ch_qr_iban(self):
+        """Check if field is populated with Swiss / Liechtenstein qr iban"""
+        for record in self:
+            if (
+                record.l10n_ch_qr_iban
+                and not record.l10n_ch_qr_iban.startswith(('CH', 'LI'))
+            ):
+                raise ValidationError(_(
+                    "Not valid Switzerland or Liechtenstein QR IBAN."))
```
yvaucher: Please use `_validate_qr_iban` instead.

And move the CH LI check in there.

It seems `_validate_qr_iban` is there for that, the constraint was missing though.

```suggestion
            if record.l10n_ch_qr_iban:
                self._validate_qr_iban(record.l10n_ch_qr_iban)
```

yvaucher (changes requested):
Would work but we can use `_validate_qr_iban` for the purpose of the constraint.

l10n_ch_qr_bill/models/res_bank.py
santostelmo: Done

l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -35,6 +41,9 @@ def _check_qr_iban_range(self, iban):
                 and 30000 <= int(iid) <= 31999)
 
     def _validate_qr_iban(self, qr_iban):
+        if qr_iban and not qr_iban.startswith(('CH', 'LI')):
+            raise ValidationError(_(
+                "Not valid Switzerland or Liechtenstein QR IBAN."))
```
yvaucher:
```suggestion
                "Not a valid Switzerland or Liechtenstein QR-IBAN."))
```

santostelmo:
@yvaucher rebased

yvaucher (approved):

PR #583, opened 2021-01-22
[FIX][l10n_ch_invoice_reports] wrong report ID to download

ajaniszewska-dev:
Fix wrong report id for download in customer portal


yvaucher (approved):
Indeed an `ir.actions.report` object is required by https://github.com/odoo/odoo/blob/13.0/addons/portal/controllers/portal.py#L382

PR #584, opened 2021-01-29
[FIX] Check presence of QR IBAN before checking validity

vrenaville:
* Purpose
We need to first check if the QR iban is filed before checking validity

https://github.com/OCA/l10n-switzerland/pull/584

### DIFF
--- a/l10n_ch_qr_bill/models/res_bank.py
+++ b/l10n_ch_qr_bill/models/res_bank.py
@@ -28,7 +28,8 @@ class ResPartnerBank(models.Model):
     def _check_ch_li_l10n_ch_qr_iban(self):
         """Validate QR-IBAN"""
         for record in self:
-            self._validate_qr_iban(record.l10n_ch_qr_iban)
+            if record.l10n_ch_qr_iban:
+                self._validate_qr_iban(record.l10n_ch_qr_iban)
 
     def _check_qr_iban_range(self, iban):
         if not iban or len(iban) < 9:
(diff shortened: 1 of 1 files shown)


ajaniszewska-dev (approved):

l10n_ch_qr_bill/models/res_bank.py
```diff
@@ -44,14 +44,15 @@ def _validate_qr_iban(self, qr_iban):
...
                 "Not a valid Switzerland or Liechtenstein QR-IBAN."))
-        # Check first if it's a valid IBAN.
-        validate_iban(qr_iban)
-        # We sanitize first so that _check_qr_iban_range()
-        # can extract correct IID from IBAN to validate it.
-        sanitized_qr_iban = sanitize_account_number(qr_iban)
-        # Now, check if it's valid QR-IBAN (based on its IID).
-        if not self._check_qr_iban_range(sanitized_qr_iban):
-            raise ValidationError(_("QR-IBAN '%s' is invalid.") % qr_iban)
+        if qr_iban:
```
carlosserratoro: You already check in the previous ```if``` if ```qr_iban``` has a value.

```suggestion
```

carlosserratoro (approved):
If possible, remove the new ```if``` while keeping the indentation, that way you save a redundant check in every call.

yvaucher:
@vrenaville I counter proposed an other fix to have `_validate...` function checking emptiness and making sure the constraint doesn't call validation for nothing.

@carlosserratoro Does it take take of your remark?

carlosserratoro:
> @carlosserratoro Does it take take of your remark?

Yes. But was just a non-blocking comment, I could live with the ```if``` :)

PR #551, opened 2020-06-11
[13.0][MIG] ebill_paynet

TDu:
This is based on #499 for 12.0 that has not been merged

- [x] account_invoice_export OCA/edi#212
- [x] base_ebill_payment_contract OCA/edi#207

PR #586, opened 2021-02-08
[13.0][ADD] ebill_paynet_account_financial_discount

TDu:
Depends on
- [ ] https://github.com/OCA/account-financial-tools/pull/1040

This module implements the discount information in the payment terms node
of the Paynet message. In regards to the module `account_financial_discount`.

When the discount information is present in the payment terms of an invoice, they
will automatically be included in the payment term of the Paynet message.

This module was included in #551 and already reviewed there. But to ease work it is now proposed on its own.

### DIFF
--- a/ebill_paynet_account_financial_discount/__init__.py
+++ b/ebill_paynet_account_financial_discount/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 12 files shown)


simahawk:
@TDu pls check the build

grindtildeath:
Depends on https://github.com/OCA/account-financial-tools/pull/1040 , so not sure we'll ever be able to merge this :disappointed:

simahawk:
@TDu  ping

TDu:
There is a potential fix for it https://github.com/OCA/l10n-switzerland/pull/623

TDu:
@simahawk This is one is red because it depends on https://github.com/OCA/account-financial-tools/pull/1040 which is being held back because of other implementation (that solve the same problem) see https://github.com/OCA/account-payment/pull/334

PR #587, opened 2021-02-18
[FIX] fix wrong prefix for tag xml_id

houssine78:
This PR brings a small fix on a wrongly prefixed account_tag which break the mis report balanced sheet


houssine78:
@fclementic2c if you can have a look it will be great.

yvaucher (approved):
Good catch

PR #588, opened 2021-02-19
[13][ADD] ebill_paynet_customer_free_ref

TDu:
This module interconnects the `sale_order_customer_free_ref` from EDI
with the `ebill_paynet` module.

The order reference in the XML messge is set with the `customer_order_number` field.
And the `customer_order_free_ref` is added as a <OTHER-REFERENCE> node.

### DIFF
--- a/ebill_paynet_customer_free_ref/models/sale_order.py
+++ b/ebill_paynet_customer_free_ref/models/sale_order.py
@@ -1,0 +1,14 @@
+# Copyright 2021 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    @api.depends("customer_order_number")
+    def _compute_paynet_client_order_ref(self):
+        for order in self:
+            order.paynet_client_order_ref = order.customer_order_number
--- a/setup/ebill_paynet_customer_free_ref/odoo/addons/ebill_paynet_customer_free_ref
+++ b/setup/ebill_paynet_customer_free_ref/odoo/addons/ebill_paynet_customer_free_ref
@@ -1,0 +1,1 @@
+../../../../ebill_paynet_customer_free_ref
--- a/ebill_paynet_customer_free_ref/__init__.py
+++ b/ebill_paynet_customer_free_ref/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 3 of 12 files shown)


setup/ebill_paynet_customer_free_ref/odoo/addons/ebill_paynet_customer_free_ref
```diff
@@ -0,0 +1 @@
+../../../../ebill_paynet_customer_free_ref
```
carlosserratoro, line 1: Aren't the contents of ```setup``` generated automatically by the tool ```black```? (same for the next file, *setup.py*)

carlosserratoro (approved):

setup/ebill_paynet_customer_free_ref/odoo/addons/ebill_paynet_customer_free_ref
simahawk: if you are adding a new module it's always better to include this in the PR because it makes possible to install the module via pip directly from the PR

ebill_paynet_customer_free_ref/messages/discount-2003A.xml.jinja
```diff
@@ -0,0 +1,8 @@
+{%- if discount %}
```
simahawk: is this wanted?

TDu: No, it was not. Removed.

sebalix (approved):

jbaudoux (approved):
LGTM

ebill_paynet_customer_free_ref/models/sale_order.py
```diff
@@ -0,0 +1,14 @@
...
+from odoo import api, models
+
+
+class SaleOrder(models.Model):
+
+    _inherit = "sale.order"
+
+    @api.depends("customer_order_number")
+    def _compute_paynet_client_order_ref(self):
+        for order in self:
```
grindtildeath, line 13: No call to super in this? Maybe add a comment to state why.

grindtildeath (approved):
LG but CI needs fixing

ebill_paynet_customer_free_ref/models/sale_order.py
TDu: No call  to super because the value retrieved by the base implementation is made obsolete by this module.

TDu:
The merge seems to fail because the `ebill_paynet` module has no README file generated ?

simahawk:
> The merge seems to fail because the `ebill_paynet` module has no README file generated ?

no, because the generation of the README fails for some bad rst. You should try to run https://github.com/OCA/maintainer-tools#readme-generator

TDu:
Yes, I was on that here we go https://github.com/OCA/l10n-switzerland/pull/623

TDu:
@simahawk Could you try again to merge. Thanks

PR #595, opened 2021-02-28
[MIG] ebill_paynet: Migration to 14.0

dzungtran89:
Dependencies:
- base_ebill_payment_contract: https://github.com/OCA/edi/pull/322 
- account_invoice_export: https://github.com/OCA/edi/pull/299


mmequignon (approved):

TDu:
@dzungtran89 Hello, that module has finally been merged in 13.0 and had a little rebasing done to it.
Could you update your  PR thank you

dzungtran89:
@TDu sure, we'll do it

dzungtran89:
Hi @TDu , I just updated the PR

ebill_paynet/tests/test_ebill_paynet.py
```diff
@@ -98,7 +114,7 @@ def test_takeShipment(self):
         attachment = self.env["ir.attachment"].search(
             [["res_model", "=", "res.country"], ["res_id", "=", ch.id]]
         )
-        attachment = self.env.ref("mail.msg_discus4_attach1")
+        attachment = self.env.ref("mass_mailing.mass_mail_attach_1")
```
TDu: It's a bummer to add a new dependency `mass_mailing` to this module only for the test.
This test only sends any file to the service and checks that it receives a shipment name.
Could you try to use another file as to not add the dependency on `mass_mailing`. Either one in Odoo or one in `ebill_paynet/tests/examples`

ebill_paynet/models/res_partner_bank.py
```diff
@@ -0,0 +1,30 @@
...
+        self,
+        amount,
+        free_communication,
+        structured_communication,
+        currency,
+        debtor_partner,
+        qr_method=None,
+        silent_errors=True,
+    ):
+        # FIXME Workaround to bypass error None type of native update
```
TDu: Thanks for the details  :+1: 
I wonder why we have this issue with this module. And if the qr code generation works sometimes.

TDu (commented):
Thanks for the migration, could you look at removing that dependency as explained in one of the comment, thank you.

ebill_paynet/models/res_partner_bank.py
dzungtran89: Hi @TDu , there is a check of eligible qr [here](https://github.com/odoo/odoo/blob/bc8e2d67b5980bb963f0e663f3c3d761ca4af169/addons/account/models/res_bank.py#L37). In the native module `l10n_ch`, it has overrided this func [here](https://github.com/odoo/odoo/blob/bc8e2d67b5980bb963f0e663f3c3d761ca4af169/addons/l10n_ch/models/res_bank.py#L289), but it just returns True for a specific case

ebill_paynet/tests/test_ebill_paynet.py
dzungtran89: well noted, I'll update it

dzungtran89:
I just synced the PR with odoo upstream, they refactored the code in this [commit](https://github.com/odoo/odoo/commit/07df98cfa8a6d94b6fea2168d169f91473a00169)

ebill_paynet/tests/test_ebill_paynet.py
```diff
@@ -114,7 +114,10 @@ def test_takeShipment(self):
         attachment = self.env["ir.attachment"].search(
             [["res_model", "=", "res.country"], ["res_id", "=", ch.id]]
         )
-        attachment = self.env.ref("mass_mailing.mass_mail_attach_1")
+        if not attachment:
```
TDu: Thanks for that change, could you also remove this `if` and the previous `attachment` assignation which is not needed anymore.

TDu (approved):
LG, I left one more comment for a change in the test. You can squash the change made into the one migration commit.
Thank you

ebill_paynet/tests/test_ebill_paynet.py
dzungtran89: ok, I pushed the update

dzungtran89:
fyi, travis and runbot are red because I removed `test-requirements.txt` file

TDu:
@dzungtran89 I had to do a fix on the readme files for the 13 version https://github.com/OCA/l10n-switzerland/pull/623
Could you include it here as well. Thank you

dzungtran89:
> 
> 
> @dzungtran89 I had to do a fix on the readme files for the 13 version #623 Could you include it here as well. Thank you

Hi @TDu , I updated the PR

simahawk:
https://github.com/OCA/edi/pull/299 is getting merged

PR #599, opened 2021-03-12
[12.0] Fix l10n_ch_base_bank account type calculation

santostelmo:
Account number is changed in `_update_acc_name` to have a structure like `parner / Postal number acc_number`. 
https://github.com/OCA/l10n-switzerland/blob/a6f79e04d9d2e8c19f5592b5fe974be19c5f721f/l10n_ch_base_bank/models/bank.py#L289
The goal is to  avoid duplicated bank accounts. But with this change, the type of the account is not recognized as postal anymore. So we consider valid if has substring "/Postal number" is found in account number

### DIFF
--- a/l10n_ch_base_bank/models/bank.py
+++ b/l10n_ch_base_bank/models/bank.py
@@ -18,6 +18,16 @@ def validate_l10n_ch_postal(postal_acc_number):
     """
     if not postal_acc_number:
         raise ValidationError(_("There is no postal account number."))
+
+    # Account number was changed in _update_acc_name to have a structure like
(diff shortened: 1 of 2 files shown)


yvaucher:
The type is not important in the modules.
I don't think it's a good idea to do this.

Postal is not correct anyway. An account with an ISR subscription number (content of `l10n_ch_postal`) can be a bank account or a postal account. I would be more in favor or removing the type `postal` which is useless.

The type field doesn't suit our use, because you could fill an IBAN in `acc_number` and an ISR subscription number in the `l10n_ch_postal`. In that case we can't put `postal` and `iban` in the type at the same time.

Thus I'm closing this.

PR #600, opened 2021-03-16
[8.0] l10n_ch_qr_bill: fixed missing QRR data in qr code

fredzamoabg:
probably during backporting of the module l10n_ch_qr_bill was keept the wrong account.invoice validate method, this causes a missing QRR reference inserted in QR code.

In this PR I removed dependency of field 'name' for QRR reference insertion in qr code, removing also unrequired account.invoice validate method.

### DIFF
--- a/l10n_ch_qr_bill/models/account_invoice.py
+++ b/l10n_ch_qr_bill/models/account_invoice.py
@@ -17,7 +17,9 @@ class AccountInvoice(models.Model):
 
     l10n_ch_qrr = fields.Char(
         compute='_compute_l10n_ch_qrr',
+        string='QRR ref',
         store=True,
+        readonly=True,
         help='The reference QRR associated with this invoice',
     )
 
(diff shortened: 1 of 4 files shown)


SimoneVagile (approved):

l10n_ch_qr_bill/models/account_invoice.py
```diff
@@ -279,13 +278,3 @@ def _validate_qrr(self):
...
             )
         return True
-
-    def action_invoice_open(self):
-        res = super(AccountInvoice, self).signal_workflow('invoice_open')
-        for rec in self:
-            # Define reference once number has been generated
-            if rec.type == "out_invoice":
-                if not rec.name and rec.partner_bank_id._is_qr_iban():
-                    rec.name = rec.l10n_ch_qrr
```
yvaucher, line 289: IIRC this was there to ensure the full workflow works for reconciliation. It most likely doesn't make sens for v8.

If you want a full reconciliation process you might need a `transaction_id` glue module for that with `base_transaction_id`. (like it was done with ISR)

yvaucher (approved):
Code LGTM

Seems there are some formatting issues in the PO file but this is mild.

PR #601, opened 2021-03-17
[10.0] l10n_ch_qr_bill: Fix _get_communications

grindtildeath:
_get_communications function defines the reference and
reference_type to be used in the generation of the QR bill.

has_qrr only checks if the reference is defined in the name
field of account.invoice. But the reference is displayed on the
report if the bank account is set up for QR bill.

This allows to generate QR XML properly even if the
name field is empty on customer invoices.

### DIFF
--- a/l10n_ch_qr_bill/models/account_invoice.py
+++ b/l10n_ch_qr_bill/models/account_invoice.py
@@ -140,7 +140,8 @@ class AccountInvoice(models.Model):
             record.l10n_ch_qrr_spaced = spaced_qrr
 
     def _get_communications(self):
-        if self.has_qrr():
+        is_qrr = self.partner_bank_id._is_qr_iban()
+        if is_qrr:
             structured_communication = self.l10n_ch_qrr
             free_communication = ''
         else:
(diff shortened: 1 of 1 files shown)


grindtildeath:
ping @yvaucher

l10n_ch_qr_bill/models/account_invoice.py
```diff
@@ -140,7 +140,8 @@ def _space_qrr(ref):
             record.l10n_ch_qrr_spaced = spaced_qrr
 
     def _get_communications(self):
-        if self.has_qrr():
+        is_qrr = self.has_qrr() or self.partner_bank_id._is_qr_iban()
```
i-vyshnevska: I don't know all the stuff in here, but on line 330 there is a check to have both of this conditions

yvaucher: To give some perspective here is the rule:

:heavy_check_mark:  QR-IBAN and QRR
:x: QR-IBAN and not QRR
:heavy_check_mark: IBAN and Free reference
:heavy_check_mark: IBAN and QRR (suspicious but considered as a Free reference thus is OK)

Which mean that we can rely solely on `self.partner_bank_id._is_qr_iban()` to know if `structured_communication` must be set or not.

yvaucher:
```suggestion
        is_qrr = self.partner_bank_id._is_qr_iban()
```

yvaucher (changes requested):
Changing to a check only on bank account `_is_qr_iban()` would be more relevent and in line with the QWeb report.

yvaucher (approved):

PR #602, opened 2021-03-17
[12.0]  l10n_ch_base_bank: Avoid breaking with =like and =ilike operators

Martronic-SA:
=ilike and =like don't exist in sql so don't add them directly to the query but replace them by ilike (code above borrowed from osv/expression.py)


yvaucher (approved):
Funny, the `=like` and `=ilike` are mentioned at line 37 but weren't tested.

Approved

PR #603, opened 2021-03-19
[13.0][FIX] l10n_ch_base_bank

TDu:
On an unrelated PR the test for l10n_ch_base_bank were broken.
It seems to be related to this new change introduced by

* https://github.com/odoo/odoo/pull/67599

Which test the valid of the type journal.

So this fixes the test, but I am not sure if they are still meaningfull.

### DIFF
--- a/l10n_ch_base_bank/tests/test_create_invoice.py
+++ b/l10n_ch_base_bank/tests/test_create_invoice.py
@@ -18,11 +18,20 @@ class TestCreateMove(common.SavepointCase):
         # define company bank account
         cls.bank_journal = cls.env["account.journal"].create(
             {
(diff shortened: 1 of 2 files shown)


yvaucher (commented):
Why is the fix in `test_search_invoice` different than in `test_create_invoice` ?

TDu:
> Why is the fix in `test_search_invoice` different than in `test_create_invoice` ?

Well both test file started different. One created a res.partner.bank and the other not.
I have change them so they prepare the environment in the same way.

TDu:
@yvaucher Do you want to have another look at this ?
It could probably fix master that is red.

yvaucher (approved):

PR #605, opened 2021-03-25
[IMP] merge report invoice per invoice to preserve order

vrenaville:
* Purpose
If we use this report in tree mode, we have all invoice report then all qr report . you will need to do a manual 
sorting after.
with this modification I merge report on the end to preserve report orderr

### DIFF
--- a/l10n_ch_invoice_reports/models/report.py
+++ b/l10n_ch_invoice_reports/models/report.py
@@ -45,26 +45,26 @@ class IrActionsReport(models.Model):
         ]
         if self.report_name not in reports or not res_ids:
             return super().render_qweb_pdf(res_ids, data)
-        inv_report = self._get_report_from_name("account.report_invoice")
-        invoice_pdf, _ = inv_report.render_qweb_pdf(res_ids, data)
-        invoice_pdf_io = io.BytesIO(invoice_pdf)
+        tmp_stream_io = []
+        for res_id in res_ids:
+            inv_report = self._get_report_from_name("account.report_invoice")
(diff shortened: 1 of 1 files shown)


l10n_ch_invoice_reports/models/report.py
```diff
@@ -45,26 +45,26 @@ def render_qweb_pdf(self, res_ids=None, data=None):
...
+        for res_id in res_ids:
+            inv_report = self._get_report_from_name("account.report_invoice")
+            invoice_pdf, _ = inv_report.render_qweb_pdf(res_id, data)
+            invoice_pdf_io = io.BytesIO(invoice_pdf)
 
-        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
-        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
-        isr_pdf_io = io.BytesIO(isr_pdf)
+            isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
+            isr_pdf, _ = isr_report.render_qweb_pdf(res_id, data)
```
yvaucher, line 55: Are we computing ISR and QR report everytime we call the merge? Even when we don't use them?

Seems we could check `self.report_name` before generating both.
(Not directly related to this PR though.

yvaucher (approved):
LGTM

Raised an issue that is not related to the current PR. Better to solve it after this one is merged.

yvaucher:
Travis failure on module `l10n_ch_base_bank` is not related to this PR.

ajaniszewska-dev (approved):

PR #610, opened 2021-08-24
[10.0][FIX] l10n_ch_base_bank

yankinmax:
At some point, while importing Invoice as PDF with Import Vendor Bill wizard, we have an issue on the search for existing Invoices.
At the moment, if the value is not set, we got:
`=like` operator transforms in Odoo to Postgres query like `SELECT column like 'value'`
this is not what we want. Instead, to avoid missing Invoices, we need this kind of query:
`like` operator transforms in Odoo to Postgres query like `SELECT column like '%svalue%s'`

### DIFF
--- a/l10n_ch_base_bank/models/invoice.py
+++ b/l10n_ch_base_bank/models/invoice.py
@@ -28,17 +28,17 @@ class AccountInvoice(models.Model):
                 continue
             if value:
                 value = value.replace(' ', '')
-                if not value:
-                    # original value contains only spaces, the query
-                    # would return all rows, so avoid a costly search
-                    # and drop the domain triplet
-                    continue
-                # add wildcards for the like search, except if the operator
-                # is =like of =ilike because they are supposed to be there yet
-                if operator.startswith('='):
-                    operator = operator[1:]
-                else:
-                    value = '%%%s%%' % (value,)
+            if not value:
+                # original value contains only spaces, the query
+                # would return all rows, so avoid a costly search
+                # and drop the domain triplet
+                continue
+            # add wildcards for the like search, except if the operator
+            # is =like of =ilike because they are supposed to be there yet
+            if operator.startswith('='):
+                operator = operator[1:]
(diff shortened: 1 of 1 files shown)


l10n_ch_base_bank/models/invoice.py
```diff
@@ -39,11 +39,11 @@ def _search(self, args, offset=0, limit=None, order=None, count=False,
...
-                     "WHERE REPLACE(reference, ' ', '') %s %%s" %
-                     (operator,))
-            self.env.cr.execute(query, (value,))
-            ids = [t[0] for t in self.env.cr.fetchall()]
+                query = ("SELECT id FROM account_invoice "
+                        "WHERE REPLACE(reference, ' ', '') %s %%s" %
+                        (operator,))
+                self.env.cr.execute(query, (value,))
+                ids = [t[0] for t in self.env.cr.fetchall()]
             domain.append(('id', 'in', ids))
```
yvaucher: Please also indent this line.

l10n_ch_base_bank/models/invoice.py
```diff
@@ -39,12 +39,12 @@ def _search(self, args, offset=0, limit=None, order=None, count=False,
                     operator = operator[1:]
                 else:
                     value = '%%%s%%' % (value,)
```
yvaucher: Change of indent must be done on the check `if not value:`

You can remove 4 space on line 31 til 41

And leave 42 to 47 to its former indent.

yvaucher (approved):

i-vyshnevska (commented):
a meaningful commit message and PR description is required, the point is maybe this fix needs to be taken to other versions so you need to describe the issue you are fixing here 
https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#commit-message

i-vyshnevska (approved):

yvaucher (approved):

PR #613, opened 2021-09-06
[12.0][ADD] module l10n_ch_delivery_carrier_label_quickpac

StephaneMangin:
Previous PR : https://github.com/OCA/delivery-carrier/pull/413
This carrier only delivers to Switzerland.

This is a new implementation for delivery carrier interaction with Quickpac.

This module has been based on  [oca/delivery-carrier/delivery_carrier_label_postlogistics](https://github.com/OCA/delivery-carrier/tree/12.0/delivery_carrier_label_postlogistics). It uses an external python library that implements the Quickpac API through its OpenAPI generated client (Swagger Codegen, see: https://pypi.org/project/quickpac/)


gurneyalex (approved):

l10n_ch_delivery_carrier_label_quickpac/data/product.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo noupdate="1">
+    <record id="product_quickpac_service" model="product.product">
+        <field name="image" type="base64" file="delivery_carrier_label_quickpac/static/img/quickpac-logo.jpg"/>
```
gurneyalex:
```suggestion
        <field name="image" type="base64" file="l10n_ch_delivery_carrier_label_quickpac/static/img/quickpac-logo.jpg"/>
```

gurneyalex: same fix to be applied in all places where the logo is referenced

StephaneMangin: Already done. Travis is restarted ^^

ajaniszewska-dev (approved):

l10n_ch_delivery_carrier_label_quickpac/models/stock.py
```diff
@@ -0,0 +1,79 @@
...
+        for picking in self:
+            partner = picking.partner_id
+            delivery_type = picking.carrier_id.delivery_type
+            if delivery_type != "quickpac" or not partner:
+                continue
+            service = QuickpacWebService(picking.company_id)
+            if not service.is_deliverable_zipcode(partner.zip):
+                raise UserError(
+                    _("This partner is undeliverable (invalid zipcode).")
+                )
```
guewen: I think it is not a good idea to block the selection of the delivery method at this point.
A given ZIP can be invalid today, but valid in 3 days (or valid today and invalid in 3 days), so if a user sets the delivery method some days before the delivery, IMO they should be able to select it (which can be a default value or set from the sale order BTW) whatever. Only the generation of the label should actually be prevented if the ZIP is not valid.

StephaneMangin: You're right, i'll remove that part. The error will raise if needed later on after all

guewen: Thanks!

yvaucher:
Question, why putting it here instead of delivery-carrier repository where the postlogistics module is (which is also only for Switzerland)

StephaneMangin:
> Question, why putting it here instead of delivery-carrier repository where the postlogistics module is (which is also only for Switzerland)

After followed this discussion https://github.com/OCA/delivery-carrier/pull/413#issuecomment-911528645

yvaucher:
Thus we might want to pull delivery_postlogitics here *sigh*

yvaucher:
is it necessary to have `l10n_ch` as a prefix? I mean the context is already there and AFAIK we don't depend on the `l10n_ch` chart of account.

l10n_ch_delivery_carrier_label_quickpac/models/delivery.py
```diff
@@ -0,0 +1,135 @@
...
+            else:
+                # Allows to set multiple optional single option in order to
+                # let the user select them
+                single_option_types = [
+                    'label_layout',
+                    'output_format',
+                    'resolution',
+                    'basic',
+                    'additional',
+                    'delivery',
```
guewen: I don't think these types should have the "single" constraint. We may need several additional services on a delivery method.

l10n_ch_delivery_carrier_label_quickpac/models/stock.py
```diff
@@ -0,0 +1,64 @@
...
+            }
+
+        labels = []
+        for item in res:
+            label = item["value"]
+            filtered_packages = packages.filtered(
+                lambda p: p.name in label['item_id']
+            )
+            if "active_test" in ctx and ctx["active_test"]:
+                # ids mismatch when testing
```
i-vyshnevska, line 47: this comment is not transparent 
"when testing" means unit tests or manual test in UI?
maybe we need to inject context={"active_test": False} in someplace before?

l10n_ch_delivery_carrier_label_quickpac/models/delivery.py
```diff
@@ -0,0 +1,132 @@
...
+        then he adds additional services.
+        """
+        option_template_obj = self.env['delivery.carrier.template.option']
+
+        for carrier in self:
+            forbidden = option_template_obj.browse()
+            domain = []
+            if carrier.delivery_type != 'quickpac':
+                domain.append(('partner_id', '=', False))
+            else:
```
santostelmo: @StephaneMangin if you add `partner_id` in the domain when the delivey type is different of quickpack, you will not get any options in the search.
```suggestion
            if carrier.delivery_type == 'quickpac':
```

l10n_ch_delivery_carrier_label_quickpac/models/delivery.py
```diff
@@ -0,0 +1,132 @@
...
+        'available_option_ids',
+        'available_option_ids.quickpac_type',
+    )
+    def _compute_allowed_options_ids(self):
+        """ Compute allowed delivery.carrier.option.
+
+        We do this to ensure the user first select a basic service. And
+        then he adds additional services.
+        """
+        option_template_obj = self.env['delivery.carrier.template.option']
```
santostelmo, line 76: @StephaneMangin missing super call
```suggestion
        option_template_obj = self.env['delivery.carrier.template.option']
        super(DeliveryCarrier, self)._compute_allowed_options_ids()
```

l10n_ch_delivery_carrier_label_quickpac/quickpac/web_service.py
```diff
@@ -0,0 +1,531 @@
...
+        )
+
+        if partner.country_id.code:
+            recipient.country = partner.country_id.code.upper()
+        if partner.street2:
+            recipient.addressSuffix = sanitize_string(partner.street2)
+
+        if partner.parent_id and partner.parent_id.name != partner_name:
+            recipient.name2 = sanitize_string(partner.parent_id.name)
+            recipient.personallyAddressed = False
```
hailangvn: This should follow quickpac package definition `lib/python3.8/site-packages/quickpac/models/recipient.py:        'personally_addressed': 'PersonallyAddressed',`.
```suggestion
            recipient.personally_addressed = False
```

l10n_ch_delivery_carrier_label_quickpac/quickpac/web_service.py
```diff
@@ -0,0 +1,531 @@
...
+            street=sanitize_string(partner.street),
+            zip=partner.zip,
+            city=partner.city,
+            e_mail=email or None,
+        )
+
+        if partner.country_id.code:
+            recipient.country = partner.country_id.code.upper()
+        if partner.street2:
+            recipient.addressSuffix = sanitize_string(partner.street2)
```
hailangvn: This should follow quickpac package definition `lib/python3.8/site-packages/quickpac/models/recipient.py: 'address_suffix': 'AddressSuffix',`.
```suggestion
            recipient.address_suffix = sanitize_string(partner.street2)
```

leemannd (approved):

l10n_ch_delivery_carrier_label_quickpac/quickpac/web_service.py
```diff
@@ -0,0 +1,531 @@
...
+        return attributes
+
+    def _prepare_notification(self, picking):
+        """ Define how and who will be notified
+
+        :param picking: a picking record
+        :return: Notification
+        """
+        communication = Communication(
+            email=picking.partner_id.email,
```
santostelmo:
```suggestion
            email=picking.partner_id.email or None,
```

l10n_ch_delivery_carrier_label_quickpac/quickpac/web_service.py
```diff
@@ -0,0 +1,531 @@
...
+            mobile=picking.partner_id.mobile
```
santostelmo:
```suggestion
            mobile=picking.partner_id.mobile or None
```

l10n_ch_delivery_carrier_label_quickpac/quickpac/web_service.py
StephaneMangin: PR https://github.com/StephaneMangin/l10n-switzerland/pull/4

l10n_ch_delivery_carrier_label_quickpac/quickpac/web_service.py
StephaneMangin: https://github.com/StephaneMangin/l10n-switzerland/pull/4

leemannd:
@StephaneMangin Is it possible to rebase?

PR #625, opened 2021-10-15
[FIX] remove bic check, not needed

vrenaville:
* Purpose
Bic check is not needed

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -69,7 +69,7 @@ class AccountPaymentOrder(models.Model):
             if bank_line.local_instrument == 'CH01':
                 # Don't set the creditor agent on BVR/CH01 payments
                 return True
-            elif not partner_bank.bank_bic:
+            elif partner_bank.acc_type != 'iban' and not partner_bank.bank_bic:
                 raise UserError(_(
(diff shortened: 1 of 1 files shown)


l10n_ch_pain_base/models/account_payment_order.py
```diff
@@ -69,13 +69,6 @@ def generate_party_agent(
             if bank_line.local_instrument == 'CH01':
                 # Don't set the creditor agent on BVR/CH01 payments
                 return True
-            elif not partner_bank.bank_bic:
```
yvaucher, line 72: This check was there for pain.001 V2

And indeed since V3 has been released the BIC and bank details are deduced from the IBAN or QR-IBAN in case of payment type 3.

However, this check is still useful in case of payment type 3 and the bank account uses a postal account number (without reference).

2021-10-18-115713_741x564_scrot

In version 10.0 we weren't properly making the difference between ESR and ESR-B.
But what we can do is to ignore this check if the type of account is `iban`

```
elif not partner_bank.bank_bic and partner_bank.acc_type != "iban":
```

yvaucher (changes requested):
Please keep the check but de-activate it for all `acc_type == 'iban'`

yvaucher (approved):

PR #629, opened 2021-11-02
FIX [12.0] Update view according to l10n_ch

Martronic-SA:
hide the two default buttons with name="isr_print"
be sure that pyaments slip buttons are inserted after Send & Print buttons as there are 2 buttons with name="action_invoice_sent"


yvaucher (approved):
LGTM

PR #630, opened 2021-11-02
ebill_paynet: Use picking name as ref in REFERENCE-DATE

mmequignon:
Customers complained that the reference should be the picking name instead of the tracking reference, in the message.


TDu (approved):

TDu:
@mmequignon could you open a forward port to 14.0 please

simahawk (approved):

PR #635, opened 2021-12-22
[13.0][IMP] ebill_paynet: round taxes

i-vyshnevska:
IMO rounding shouldn't be done on jinja template data should be corrected on invoice level but as the other part implemented in this way we follow common approach

### DIFF
--- a/ebill_paynet/messages/invoice-2003A.xml
+++ b/ebill_paynet/messages/invoice-2003A.xml
@@ -169,17 +169,17 @@
       {%- endif %}
       {% endfor %}
       <Quantity Type="47" Units="PCE">{{ line.quantity }}</Quantity>
-      <Price Type="YYY">{{ line.price_subtotal|round(2) }}</Price>
-      <Price Type="AAA">{{ line.price_subtotal|round(2) }}</Price>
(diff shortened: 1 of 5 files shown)


sebalix (commented):
pre-commit is failing

ebill_paynet/messages/invoice-2003A.xml
```diff
@@ -210,7 +210,7 @@
       {%- else %}
       <Rate Category="E">0</Rate>
       {%- endif %}
-      <Amount Currency="{{ invoice.currency_id.name }}">{{ taxgroup[1] }}</Amount>
+      <Amount Currency="{{ invoice.currency_id.name }}">{{ taxgroup[1]|round(2) }}</Amount>
```
simahawk: can't we pass a function to handle this instead?

sebalix (approved):

simahawk (approved):

simahawk:
@i-vyshnevska can you write a better commit msg? eg `ebille_paynet: round taxes` or similar? FTR https://www.odoo.com/documentation/15.0/developer/misc/other/guidelines.html#commit-message-structure

TDu (approved):

PR #636, opened 2022-01-04
[FIX] set naming in accordance to the rest of implementation

i-vyshnevska:
set selection with capital letters to avoid problems with further implementation https://github.com/OCA/l10n-switzerland/blob/7da6954fdc454eb220904b9493839f0cc66e7e5d/l10n_ch_qr_bill/models/account_invoice.py#L76

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_line.py
+++ b/l10n_ch_pain_base/models/account_payment_line.py
@@ -16,5 +16,7 @@ class AccountPaymentLine(models.Model):
     def invoice_reference_type2communication_type(self):
         res = super(AccountPaymentLine, self).\
             invoice_reference_type2communication_type()
-        res.update({'bvr': 'bvr', 'qrr': 'qrr'})
+        # inconsistent naming is done intentionally to be compatible with
+        # implementation from more recent version
+        res.update({'bvr': 'bvr', 'QRR': 'qrr'})
         return res


l10n_ch_pain_base/models/account_payment_line.py
```diff
@@ -16,5 +16,5 @@ class AccountPaymentLine(models.Model):
     def invoice_reference_type2communication_type(self):
         res = super(AccountPaymentLine, self).\
             invoice_reference_type2communication_type()
-        res.update({'bvr': 'bvr', 'qrr': 'qrr'})
+        res.update({'bvr': 'bvr', 'QRR': 'qrr'})
```
mmequignon, line 21: Is it normal that `bvr` isn't capitalized as well ?

i-vyshnevska: yes it is inconsistent, QRR backported from more recent versions and this already merged so fewer changes in code

mmequignon: > yes it is inconsistent, QRR backported from more recent versions and this already merged so fewer changes in code

Ok, could you please write a `TODO` comment then, explaining that it should be unified in the future?

I pre-approve.

mmequignon (approved):

l10n_ch_pain_base/models/account_payment_line.py
i-vyshnevska: comment added but that is not TODO as we will not refactor it and on newest versions is already implemented in another way so this just for supporting current version

ajaniszewska-dev (approved):

yvaucher (approved):

PR #639, opened 2022-01-13
ebill_paynet: Add http timeout

TDu:


ebill_paynet/models/paynet_service.py
```diff
@@ -53,6 +53,10 @@ class PaynetService(models.Model):
         readonly=True,
     )
     active = fields.Boolean(default=True)
+    operation_timeout = fields.Integer(
+        string="HTTP Timeout",
+        help="Timeout for each HTTP (GET, POST) request in seconds."
```
sebalix: Does it deserve a default value to not have a blocked Odoo worker indefinitely?

ebill_paynet/models/paynet_service.py
```diff
@@ -62,13 +66,20 @@ def _compute_url(self):
             else:
                 record.url = SYSTEM_PROD_URL
 
+    def _get_dws_connection(self):
+        return PayNetDWS(
+            self.url,
+            self.use_test_service,
+            self.operation_timeout or None
```
mmequignon:
```suggestion
            operation_timeout=self.operation_timeout or None
```
This is a named argument.

mmequignon (approved):

ebill_paynet/models/paynet_service.py
TDu: Maybe but how much is safe ?

TDu: Ok, I have set one

sebalix (approved):

ebill_paynet/models/paynet_service.py
sebalix: This code already works like this too. What makes a keyword/named argument is the usage, not the declaration of the function/method. All parameters here could be named arguments, or positional ones (like in this case).
```python
PayNetDWS(operation_timetout=10, test_service='test', url='test')    # named arguments
# same than
PayNetDWS('test', 'test', 10)    # positional arguments
```

mmequignon: Yes I know, that's why I approved the PR anyway.

yvaucher (approved):

PR #646, opened 2022-01-27
[10.0][FIX] set reference_type to QRR

i-vyshnevska:
set reference to qrr as otherwise, it will propagate wrong type to order payments

### DIFF
--- a/l10n_ch_qr_bill/models/account_invoice.py
+++ b/l10n_ch_qr_bill/models/account_invoice.py
@@ -80,12 +80,11 @@ class AccountInvoice(models.Model):
         self.ensure_one()
         return ''
 
-    @api.onchange('reference')
+    @api.onchange('reference', 'partner_bank_id')
     def onchange_reference_name(self):
         # onchange_reference method defined in l10n_ch_base_bank
         # to not bring dependency on this addon we defined method which will
-        # disable behavior of onchange_reference
-        if self.has_qrr():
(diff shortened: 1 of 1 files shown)


mmequignon (approved):

l10n_ch_qr_bill/models/account_invoice.py
```diff
@@ -80,6 +80,15 @@ def _get_qrr_prefix(self):
         self.ensure_one()
         return ''
 
+    @api.onchange('reference')
+    def onchange_reference_name(self):
+        # onchange_reference method defined in l10n_ch_base_bank
```
TDu, line 85: In v10 I see a `onchange_reference` but not an `onchange_reference_name` https://github.com/OCA/l10n-switzerland/blob/7da6954fdc454eb220904b9493839f0cc66e7e5d/l10n_ch_base_bank/models/invoice.py#L60
Is this related to another pr ?

i-vyshnevska: this is the point, as both of these methods is about onchange of "reference" only one will be triggered 
the solution is not the best one, but I don't want to bring dependency, without dependency ORM is not built properly so we cannot inherit onchange_reference

TDu: But if there is no dependencies between module and both module are installed, is it reliable that the onchange are always called in the same order ?

PR #651, opened 2022-04-27
[IMP] l10n_ch_sepa: manage QRR reference_type generating XML for SEPA payment

fredzamoabg:
cover missing QRR reference_type management


stevech091 (approved):
Functional test ok

PR #652, opened 2022-05-09
[13.0] Add module l10n_ch_qr_no_amount

grindtildeath:
This module allows to print QR bills without any amount, so that the customer can define the amount to pay.

Not setting the amount allows to use the same QR code for multiple payments referencing the same invoice.


mmequignon (approved):

sebalix (approved):

ajaniszewska-dev:
@grindtildeath  fix pre-commit

l10n_ch_qr_no_amount/models/res_partner_bank.py
```diff
@@ -0,0 +1,23 @@
...
+        self, amount, currency_name, debtor_partner, reference_type, reference, comment
+    ):
+        res = super()._prepare_swiss_code_url_vals(
+            amount, currency_name, debtor_partner, reference_type, reference, comment
+        )
+        if self.env.context.get("_no_amount_qr_code", False):
+            res = self._remove_swiss_code_amount(res)
+        return res
+
+    def _remove_swiss_code_amount(self, values_list):
```
TDu, line 20:
```suggestion
    @api.model
    def _remove_swiss_code_amount(self, values_list):
```

l10n_ch_qr_no_amount/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module allows to print QR bills without any amount, so that the customer can define the amount to pay.
+
+Not setting the amount allows to use the same QR code for multiple payments referencing the same invoice.
```
TDu, line 4: Maybe could be stated that how it is done ? with the context key ?

TDu (commented):
Do we want a unit test ?

l10n_ch_qr_no_amount/models/res_partner_bank.py
grindtildeath: As it's not done in the super methods, I guess we can avoid adding it if unnecessary:
https://github.com/odoo/odoo/blob/13.0/addons/l10n_ch/models/res_bank.py#L192
https://github.com/odoo/odoo/blob/13.0/addons/l10n_ch/models/res_bank.py#L264

l10n_ch_qr_no_amount/readme/DESCRIPTION.rst
grindtildeath: I added a bit of explanation how to use it. I don't think it makes sense to explain the technical side in the module's description.

grindtildeath:
@TDu I guess we don't need a unit test to only check if a value was removed from the list or do we?

TDu:
@grindtildeath  Thank you, LG

PR #656, opened 2022-06-27
[13.0][IMP] l10n_ch_invoice_reports: generate reports based on settings

ajaniszewska-dev:
- Added settings options to control which report we want to generate.
- Updated email.template generate_email() method to use invoice report linked as optional attachment.


l10n_ch_invoice_reports/models/res_company.py
```diff
@@ -0,0 +1,12 @@
...
+
+class ResCompany(models.Model):
+    _inherit = "res.company"
+
+    print_qr_invoice = fields.Boolean(
+        string="Print invoice with QR bill", readonly=False,
+    )
+    print_isr_invoice = fields.Boolean(
+        string="Print invoice with ISR payslip", readonly=False,
+    )
```
ivantodorovich:
```suggestion
    print_qr_invoice = fields.Boolean(string="Print invoice with QR bill")
    print_isr_invoice = fields.Boolean(string="Print invoice with ISR payslip")
```

by default, fields aren't readonly

ivantodorovich (approved):
LG, but pre-commit is failing.

From the error msg it seems unrelated. Looks like this repo could use an oca repository template update (`copier update`)

Also, would be nice to have some unit tests. Pretty big overrides there

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,58 @@
...
+            related_model = self.env[self.model_id.model].browse(res_id)
+
+            if related_model._name == "account.move":
+                rslt[res_id]["attachments"] = False
+                template = res_ids_to_templates[res_id]
+                self._render_template(template.report_name, template.model, res_id)
+                new_attachments = []
+                if (
+                    self.env.company.print_qr_invoice
+                    or self.env.company.print_isr_invoice
```
i-vyshnevska: better to rely on invoice object than on self.env.company

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,29 @@ def merge_pdf_in_memory(self, docs):
...
-        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
-        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
-        isr_pdf_io = io.BytesIO(isr_pdf)
+        reports_list = []
+        for invoice in res_ids:
+            inv_report = self._get_report_from_name("account.report_invoice")
+            invoice_pdf, _ = inv_report.render_qweb_pdf(invoice, data)
+            reports_list.append(invoice_pdf)
+
+            if self.env.company.print_qr_invoice:
```
i-vyshnevska: better to rely on invoice object than on self.env.company

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,58 @@
...
+                ):
+                    # We add an attachment containing the invoice + QR/ISR payslip
+                    # depending on option chosen in settings
+                    invoice_payslip_report_name = (
+                        "invoice_" + (related_model.name) + "_with_payslip.pdf"
+                    )
+                    invoice_payslip_pdf = self.env.ref(
+                        "l10n_ch_invoice_reports.account_move_payment_report"
+                    ).render_qweb_pdf([res_id])[0]
```
i-vyshnevska: can we here relly on report linked to the template instead of hardcoded name?

i-vyshnevska (commented):
better to keep commits separately as one is already merged and that is backporting and the other is a new feature we may need to forward port

i-vyshnevska (approved):

l10n_ch_invoice_reports/data/reports.xml
```diff
@@ -8,7 +8,7 @@
         name="l10n_ch_invoice_reports.account_move_payment_report"
         file="l10n_ch_invoice_reports.account_move_payment_report"
         attachment_use="True"
-        attachment="invoice_'+(object.name)+'_with_payslip.pdf"
```
ajaniszewska-dev: object.name wasn't correctly replaced by invoice name on mail.template attachment, decided to update  it in model.

l10n_ch_invoice_reports/data/reports.xml
```diff
@@ -1,39 +1,15 @@
...
-        string="Invoice with ISR"
-        name="l10n_ch_invoice_reports.invoice_isr_report_main"
-        file="l10n_ch_invoice_reports.invoice_isr_report_main"
-        attachment_use="False"
-        attachment=""
+        string="Invoice with payment slip(s)"
+        name="l10n_ch_invoice_reports.account_move_payment_report"
+        file="l10n_ch_invoice_reports.account_move_payment_report"
+        attachment_use="True"
+        attachment="invoice_with_payslip.pdf"
```
i-vyshnevska:
```suggestion
        attachment="invoice_'+(object.name or '')+'_with_payslip.pdf"
```
name assigned when invoice confirmed so sometimes it can be empty

ajaniszewska-dev: @i-vyshnevska, even with this  on posted invoice:
image


btw. can be attachment's name translated? they would like to have it in different languages depending on customer...

i-vyshnevska: "'invoice_'+(object.name or '')+'_with_payslip.pdf'"
quotes

ajaniszewska-dev: this string isn't taken properly, doesn't matter what i am doing. Maybe there is some trick with report action (which we dont call)
image

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,60 @@
...
+        if isinstance(res_ids, int):
+            res_ids = [res_ids]
+            multi_mode = False
+
+        res_ids_to_templates = self.get_email_template(res_ids)
+        for res_id in res_ids:
+            related_model = self.env[self.model_id.model].browse(res_id)
+
+            if related_model._name == "account.move":
+                related_model.attachment_ids = False
```
i-vyshnevska: we cannot zero all attachments belonging to the model here, that can be not reported as just attachments loaded by the user the reports created in the super method should be in rslt[res_id]["attachments"], they can be removed

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,58 @@
...
+                # We add an optional attachment from mail template (if set)
+                if self.report_template:
+                    report_name = self.report_template.attachment
+                    if (
+                        self.report_template.report_name
+                        == "l10n_ch_invoice_reports.account_move_payment_report"
+                    ):
+                        report_name = _(
+                            "invoice_%s_with_payslip.pdf"
+                        ) % related_model.name.replace("/", "_")
```
ajaniszewska-dev, line 41: this takes translation based on lang in context (UI lang). not the customer lang.

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,60 @@
...
+        if isinstance(res_ids, int):
+            res_ids = [res_ids]
+            multi_mode = False
+
+        res_ids_to_templates = self.get_email_template(res_ids)
+        for res_id in res_ids:
+            related_model = self.env[self.model_id.model].browse(res_id)
+
+            if related_model._name == "account.move":
+                rslt[res_id]["attachments"] = False
```
i-vyshnevska: attachments still will be created and linked to account.move we probably need to check if ids there and unlink them

ajaniszewska-dev: we could do something like that, to be sure we remove only attachments generated earlier by odoo (afais there is possibility to add manually other attachments to invoice)

```
attachments_to_delete = related_model.attachment_ids.filtered(
                    lambda x: x.name.startswith('ISR') 
                    or x.name.startswith('QR') 
                    or x.name.startswith(x.res_name.replace('/','_'))
                )
                attachments_to_delete.unlink()
```
                
need to test if this helps.

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,45 @@
...
+        "invoice_partner_bank_id.l10n_ch_isr_subscription_eur",
+        "invoice_partner_bank_id.l10n_ch_isr_subscription_chf",
+    )
+    @api.depends_context("_mail_template_no_attachments")
+    def _compute_l10n_ch_isr_valid(self):
+        """Returns True if all the data required to generate the ISR are present"""
+        for record in self:
+            if self.env.context.get("_mail_template_no_attachments", False):
+                record.l10n_ch_isr_valid = False
+                return False
```
TDu: This `return` statement should not be there. Because all records will not be computed.

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    def generate_email(self, res_ids, fields=None):
+        """ Method overridden in order remove QR/ISR payslips and use
+            Invoice report with payslip generated by this module
+        """
+        self = self.with_context(_mail_template_no_attachments=True)
```
TDu: Could this function be simplify by first checking if the model is "account.move" or not ?

l10n_ch_invoice_reports/models/account_invoice.py
simahawk: Indeed, use `continue`

l10n_ch_invoice_reports/models/mail_template.py
simahawk: Agree. Also this seems wrong because you are setting this ctx key every time.
We should set it - if possible - where the email generation is triggered.

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,61 @@
...
+        rslt = super(MailTemplate, self).generate_email(res_ids, fields)
+
+        multi_mode = True
+        if isinstance(res_ids, int):
+            res_ids = [res_ids]
+            multi_mode = False
+
+        res_ids_to_templates = self.get_email_template(res_ids)
+        for res_id in res_ids:
+            related_model = self.env[self.model_id.model].browse(res_id)
```
simahawk: this is not the model, is the record

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,61 @@
...
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    def generate_email(self, res_ids, fields=None):
+        """ Method overridden in order remove QR/ISR payslips and use
+            Invoice report with payslip generated by this module
+        """
+        self = self.with_context(_mail_template_no_attachments=True)
+        rslt = super(MailTemplate, self).generate_email(res_ids, fields)
```
simahawk:
```suggestion
        result = super().generate_email(res_ids, fields)
```

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,61 @@
...
+            res_ids = [res_ids]
+            multi_mode = False
+
+        res_ids_to_templates = self.get_email_template(res_ids)
+        for res_id in res_ids:
+            related_model = self.env[self.model_id.model].browse(res_id)
+
+            if related_model._name == "account.move":
+                rslt[res_id]["attachments"] = False
+                template = res_ids_to_templates[res_id]
```
simahawk: the whole code here should be moved to its own method

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,61 @@
...
+                        0
+                    ]
+                    report_pdf = base64.b64encode(report_pdf)
+                    new_attachments.append((report_name, report_pdf))
+
+                attachments_list = (
+                    multi_mode
+                    and rslt[res_id].get("attachments", False)
+                    or rslt.get("attachments", False)
+                )
```
simahawk:
```suggestion
                attachments_list = (
                    result[res_id].get("attachments", []) if multi_mode else result.get("attachments", [])
                )
```

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
-        inv_report = self._get_report_from_name("account.report_invoice")
-        invoice_pdf, _ = inv_report.render_qweb_pdf(res_ids, data)
-        invoice_pdf_io = io.BytesIO(invoice_pdf)
 
-        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
-        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
-        isr_pdf_io = io.BytesIO(isr_pdf)
+        reports_list = []
+        for invoice in res_ids:
+            inv_obj = self.env["account.move"].browse(invoice)
```
simahawk: browse them all before

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
             return super().render_qweb_pdf(res_ids, data)
-        inv_report = self._get_report_from_name("account.report_invoice")
-        invoice_pdf, _ = inv_report.render_qweb_pdf(res_ids, data)
-        invoice_pdf_io = io.BytesIO(invoice_pdf)
 
-        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
-        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
-        isr_pdf_io = io.BytesIO(isr_pdf)
+        reports_list = []
+        for invoice in res_ids:
```
simahawk:
```suggestion
        for inv_id in res_ids:
```

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
+            inv_obj = self.env["account.move"].browse(invoice)
+            inv_report = self._get_report_from_name("account.report_invoice")
```
simahawk: do not call this on each iteration

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
+            invoice_pdf, _ = inv_report.render_qweb_pdf(invoice, data)
+            reports_list.append(invoice_pdf)
+
+            if inv_obj.company_id.print_qr_invoice:
+                qr_report = self._get_report_from_name("l10n_ch.qr_report_main")
```
simahawk: same

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
+                qr_pdf, _ = qr_report.render_qweb_pdf(invoice, data)
+                reports_list.append(qr_pdf)
 
-        qr_report = self._get_report_from_name("l10n_ch.qr_report_main")
-        qr_pdf, _ = qr_report.render_qweb_pdf(res_ids, data)
-        qr_pdf_io = io.BytesIO(qr_pdf)
+            if inv_obj.company_id.print_isr_invoice:
+                isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
```
simahawk: same

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
-        elif self.report_name == reports[1]:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, qr_pdf_io])
-        else:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io, qr_pdf_io])
-        invoice_pdf_io.close()
-        isr_pdf_io.close()
-        qr_pdf_io.close()
+        io_list = [io.BytesIO(pdf_file) for pdf_file in reports_list]
+        pdf = self.merge_pdf_in_memory(io_list)
+        [io_file.close() for io_file in io_list]
```
simahawk: this is creating a list that we don't need

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
-        if self.report_name == reports[0]:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io])
-        elif self.report_name == reports[1]:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, qr_pdf_io])
-        else:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io, qr_pdf_io])
-        invoice_pdf_io.close()
-        isr_pdf_io.close()
-        qr_pdf_io.close()
+        io_list = [io.BytesIO(pdf_file) for pdf_file in reports_list]
```
simahawk: why don't you append `BytesIO` directly instead of looping again?

simahawk (changes requested):
Commits to be reworded too.

l10n_ch_invoice_reports/models/report.py
ajaniszewska-dev: then iteration over the elements and closing one by one would be better? Don't see better option.

simahawk: yes

l10n_ch_invoice_reports/models/mail_template.py
ajaniszewska-dev: @simahawk @TDu  reworked method and overwritten ctx in action for that, should be better now.

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+        multi_mode = True
+        if isinstance(res_ids, int):
+            res_ids = [res_ids]
+            multi_mode = False
+
+        res_ids_to_templates = self.get_email_template(res_ids)
+        for res_id in res_ids:
+            related_record = self.env[self.model_id.model].browse(res_id)
+            template = res_ids_to_templates[res_id]
+            if related_record._name == "account.move":
```
simahawk: you know the model before, don't you? This check should happen after the call to super and return the result straight if not satisfied.

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,32 @@ def merge_pdf_in_memory(self, docs):
...
         reports = [
-            "l10n_ch_invoice_reports.invoice_isr_report_main",
-            "l10n_ch_invoice_reports.invoice_qr_report_main",
-            "l10n_ch_invoice_reports.invoice_qr_isr_report_main",
+            "l10n_ch_invoice_reports.account_move_payment_report",
         ]
+
+        inv_report = self._get_report_from_name("account.report_invoice")
+        qr_report = self._get_report_from_name("l10n_ch.qr_report_main")
+        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
```
TDu, line 50: These 3 lines should be below the next `if` case so they are not run for every kind of reports.

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,32 @@ def merge_pdf_in_memory(self, docs):
...
-        if self.report_name == reports[0]:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io])
-        elif self.report_name == reports[1]:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, qr_pdf_io])
-        else:
-            pdf = self.merge_pdf_in_memory([invoice_pdf_io, isr_pdf_io, qr_pdf_io])
-        invoice_pdf_io.close()
-        isr_pdf_io.close()
-        qr_pdf_io.close()
+        pdf = self.merge_pdf_in_memory(io_list)
```
TDu: If multiple invoices are printed at the same time, they will all be merged into the same pdf ! Is that what we want ?

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,32 @@ def merge_pdf_in_memory(self, docs):
...
             return super().render_qweb_pdf(res_ids, data)
-        inv_report = self._get_report_from_name("account.report_invoice")
-        invoice_pdf, _ = inv_report.render_qweb_pdf(res_ids, data)
-        invoice_pdf_io = io.BytesIO(invoice_pdf)
 
-        isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
-        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
-        isr_pdf_io = io.BytesIO(isr_pdf)
+        io_list = []
+        self.env["account.move"].browse(res_ids)
```
TDu: Why not use the records returned and remove the search in the next `for` loop ?

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+    @api.depends(
+        "type",
+        "name",
+        "currency_id.name",
+        "invoice_partner_bank_id.l10n_ch_isr_subscription_eur",
+        "invoice_partner_bank_id.l10n_ch_isr_subscription_chf",
+    )
+    @api.depends_context("_mail_template_no_attachments")
+    def _compute_l10n_ch_isr_valid(self):
+        """Returns True if all the data required to generate the ISR are present"""
```
simahawk: does not return anything (as required)

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+        for record in self:
+            if self.env.context.get("_mail_template_no_attachments", False):
```
simahawk: What's the desired impact of this change? Can you point me to where this flag is used in this report process?

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+            self.partner_id,
+            self.invoice_payment_ref,
+        )
+
+    def action_invoice_sent(self):
+        # override to update context with new key
+        action = super().action_invoice_sent()
+        ctx = action["context"].copy()
+        ctx.update({"_mail_template_no_attachments": True})
+        action["context"] = ctx
```
simahawk: ok, now that this is in the right context, you can rename it to something meaningful, like `invoice_report_no_attachment` or something similar

```suggestion
        action = super().action_invoice_sent()
        action["context"] = dict(action.get("context", {}), invoice_report_no_attachment=True)
```

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+import base64
+
+from odoo import _, models
+
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    def generate_email(self, res_ids, fields=None):
+        """ Method overridden in order remove QR/ISR payslips and use
```
simahawk: this is a comment

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+                ) % related_record.name.replace("/", "_")
+            report_xml_id = self.report_template.xml_id
+            report_pdf = self.env.ref(report_xml_id).render_qweb_pdf([res_id])[0]
+            report_pdf = base64.b64encode(report_pdf)
+            new_attachments.append((report_name, report_pdf))
+        attachments_list = (
+            multi_mode
+            and result[res_id].get("attachments", False)
+            or result.get("attachments", False)
+        )
```
simahawk:
```suggestion
        attachments_list = (
            result[res_id].get("attachments", [])
            if multi_mode
            else result.get("attachments", [])
        )
```

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+        new_attachments = []
+        # We add an optional attachment from mail template (if set)
+        if self.report_template:
+            report_name = self.report_template.attachment
+            if (
+                self.report_template.report_name
+                == "l10n_ch_invoice_reports.account_move_payment_report"
+            ):
+                # force translation of attachment's name according to partner's lang
+                self = self.with_context(lang=related_record.partner_id.lang)
```
simahawk: do not change self, use a new var (unless translations for some reason won't work)

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+                report_name = _(
+                    "invoice_%s_with_payslip.pdf"
+                ) % related_record.name.replace("/", "_")
+            report_xml_id = self.report_template.xml_id
+            report_pdf = self.env.ref(report_xml_id).render_qweb_pdf([res_id])[0]
+            report_pdf = base64.b64encode(report_pdf)
+            new_attachments.append((report_name, report_pdf))
+        attachments_list = (
+            multi_mode
+            and result[res_id].get("attachments", False)
```
simahawk: at line 37 you reset `result[res_id]["attachments"]` to False... how can you get some here? :thinking:

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+            or result.get("attachments", False)
+        )
+        if attachments_list:
+            attachments_list.extend(new_attachments)
+        else:
+            result[res_id]["attachments"] = new_attachments
```
simahawk: Why don't you check for multi mode here?
Also, what do you do w/ attachment_list from here?

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,32 @@ def merge_pdf_in_memory(self, docs):
...
+        for inv_id in res_ids:
+            inv_obj = self.env["account.move"].search([("id", "=", inv_id)])
```
simahawk: why do you search? You have IDs and model: just browse:

```
for inv in self.env["account.move"].browse(res_ids):
....
```

simahawk (changes requested):

l10n_ch_invoice_reports/models/mail_template.py
TDu: This should be at the beginning of the function like this:
```suggestion
            if self._name == "account.move":
```
So the previous code is not run for non 'account.move' recordset.

l10n_ch_invoice_reports/models/account_invoice.py
ajaniszewska-dev: @simahawk  this was copy paste from original code, i suppose this comment is wrong as we only set flag here.

l10n_ch_invoice_reports/models/mail_template.py
ajaniszewska-dev: i don't get how this new var is supposed to be used at later stage, so we can force to change lang in context with it?

l10n_ch_invoice_reports/models/mail_template.py
ajaniszewska-dev: correct me if i am wrong:
we call this method in iteration, every time for single res_id from [res_ids] list -   afaiu multi mode is when we process multiple invoices - what's the point with checking for that here? i think we don't need this attachment list at all, we should just care about result[res_id][attachment] content. 
Majority of this code is copied from other project to not have to discover the wheel once again, could be that I missed some part of the problem here.

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+        for record in self:
+            if self.env.context.get("invoice_report_no_attachment", False):
+                record.l10n_ch_isr_valid = False
+                continue
+            record.l10n_ch_isr_valid = (
+                record.type == "out_invoice"
+                and record.name
+                and record.l10n_ch_isr_subscription
+                and record.l10n_ch_currency_name in ["EUR", "CHF"]
+            )
```
i-vyshnevska:
```suggestion
    @api.depends_context("invoice_report_no_attachment")
    def _compute_l10n_ch_isr_valid(self):
        if self.env.context.get("invoice_report_no_attachment", False):
            self.update({"l10n_ch_isr_valid": False})
        else:
            super()._compute_l10n_ch_isr_valid()
```

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,53 @@
...
+        ).inherit_id and self.invoice_partner_bank_id.validate_swiss_code_arguments(
+            self.invoice_partner_bank_id.currency_id,
+            self.partner_id,
+            self.invoice_payment_ref,
+        )
+
+    def action_invoice_sent(self):
+        # override to update context with new key
+        action = super().action_invoice_sent()
+        super().action_invoice_sent()
```
i-vyshnevska: twise?

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,64 @@
...
+
+            res_ids_to_templates = self.get_email_template(res_ids)
+            for res_id in res_ids:
+                related_record = self.env[self.model_id.model].browse(res_id)
+                template = res_ids_to_templates[res_id]
+                self.generate_invoice_attachment(
+                    result, template, res_id, related_record, multi_mode
+                )
+            return result
+        else:
```
i-vyshnevska:
```suggestion
```

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class AccountInvoiceSend(models.TransientModel):
+    _inherit = "account.invoice.send"
+
+    def send_and_print_action(self):
+        # override to update context with new key
+        self = self.with_context(invoice_report_no_attachment=True)
+        result = super().send_and_print_action()
+        return result
```
i-vyshnevska:
```suggestion
        return super().send_and_print_action()
```

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    # override to add context key dependency and handle it properly inside method
+    @api.depends_context("invoice_report_no_attachment")
+    def _compute_l10n_ch_isr_valid(self):
+        if self.env.context.get("invoice_report_no_attachment", False):
+            self.update({"l10n_ch_isr_valid": False})
```
simahawk, line 16: I think you did not answer one of my prev questions: where this flag needs to be false when printing the report?

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,48 @@
...
+        return action
+
+
+class AccountInvoiceSend(models.TransientModel):
+    _inherit = "account.invoice.send"
+
+    def send_and_print_action(self):
+        # override to update context with new key
+        self = self.with_context(invoice_report_no_attachment=True)
+        return super().send_and_print_action()
```
simahawk:
```suggestion
        return super(AccountInvoiceSend, self.with_context(invoice_report_no_attachment=True)).send_and_print_action()
```

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,62 @@
...
+
+class MailTemplate(models.Model):
+    _inherit = "mail.template"
+
+    def generate_email(self, res_ids, fields=None):
+        # Method overridden in order remove QR/ISR payslips and use
+        # Invoice report with payslip generated by this module
+
+        result = super().generate_email(res_ids, fields)
+        if self._name == "account.move":
```
simahawk: `self._name` CANNOT be `account.move`
We don't have tests for this thing, right? :sweat_smile: 

What about using `self.model_id.model`?

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,62 @@
...
+            multi_mode = True
+            if isinstance(res_ids, int):
+                res_ids = [res_ids]
+                multi_mode = False
+
+            res_ids_to_templates = self.get_email_template(res_ids)
+            for res_id in res_ids:
```
simahawk: again, do not browse 1 by 1. Browse them out of the loop.

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,62 @@
...
+                self.generate_invoice_attachment(
+                    result, template, res_id, related_record, multi_mode
+                )
+        return result
+
+    def generate_invoice_attachment(
+        self, result, template, res_id, related_record, multi_mode
+    ):
+        result[res_id]["attachments"] = False
+        self._render_template(template.report_name, template.model, res_id)
```
simahawk: am I wrong or the call to `get_email_template` already rendered the template before https://github.com/odoo/odoo/blob/13.0/addons/mail/models/mail_template.py#L317 ?
Also you are not using at all the result here...  :thinking:

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,62 @@
...
+        new_attachments = []
+        # We add an optional attachment from mail template (if set)
+        if self.report_template:
+            report_name = self.report_template.attachment
+            if (
+                self.report_template.report_name
```
simahawk: this check is repeated for all records while you could pass `add_payment_report=True` as a param to this new method

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,62 @@
...
+                == "l10n_ch_invoice_reports.account_move_payment_report"
+            ):
+                # force translation of attachment's name according to partner's lang
+                template = self.with_context(lang=related_record.partner_id.lang)
```
simahawk: this `template` var is for what?
Also, the `get_email_template` is setting the lang already...

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,62 @@
...
+                report_name = _(
+                    "invoice_%s_with_payslip.pdf"
+                ) % related_record.name.replace("/", "_")
+            report_xml_id = self.report_template.xml_id
+            report_pdf = self.env.ref(report_xml_id).render_qweb_pdf([res_id])[0]
+            report_pdf = base64.b64encode(report_pdf)
+            new_attachments.append((report_name, report_pdf))
+        attachments_list = (
+            result[res_id].get("attachments", [])
```
simahawk: `attachments` here will always be empty, see below

l10n_ch_invoice_reports/models/mail_template.py
simahawk: ok for `multi_mode` but then:
1. why are you computing multi_mode above if you don't need it at all?
2. the very 1st thing you do in this new method is to wipe the values of `result[res_id]["attachment"]` -> how can you have values here?
3. to make things more clear `generate_invoice_attachment` should return the NEW result for each res_id instead of messing around w/ the old ones. Also, not all the params seems relevant.

> Majority of this code is copied from other project to not have to discover the wheel once again, could be that I missed some part of the problem here.

Can you share this w/ me in private please?

l10n_ch_invoice_reports/models/report.py
```diff
@@ -39,32 +39,30 @@ def merge_pdf_in_memory(self, docs):
...
+        qr_report = self._get_report_from_name("l10n_ch.qr_report_main")
         isr_report = self._get_report_from_name("l10n_ch.isr_report_main")
-        isr_pdf, _ = isr_report.render_qweb_pdf(res_ids, data)
-        isr_pdf_io = io.BytesIO(isr_pdf)
 
-        qr_report = self._get_report_from_name("l10n_ch.qr_report_main")
-        qr_pdf, _ = qr_report.render_qweb_pdf(res_ids, data)
-        qr_pdf_io = io.BytesIO(qr_pdf)
+        io_list = []
+        for inv_id in self.env["account.move"].browse(res_ids):
```
simahawk:
```suggestion
        for inv in self.env["account.move"].browse(res_ids):
```

simahawk (changes requested):

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -14,49 +14,33 @@ def generate_email(self, res_ids, fields=None):
...
-            report_xml_id = self.report_template.xml_id
-            report_pdf = self.env.ref(report_xml_id).render_qweb_pdf([res_id])[0]
-            report_pdf = base64.b64encode(report_pdf)
-            new_attachments.append((report_name, report_pdf))
-        attachments_list = (
-            result[res_id].get("attachments", [])
-            if multi_mode
-            else result.get("attachments", [])
+        # force translation of attachment's name according to partner's lang
+        self = self.with_context(lang=related_record.partner_id.lang)
```
ajaniszewska-dev: @simahawk  tried to keep this template_obj = self.with_context(lang=related_record.partner_id.lang)
and then use template_obj instead of self in lines below but doesn't work (or i misunderstood something here). Anyway - pdb shows lang in context  = en_US which is probably admin user lang, not the invoice partner lang. if you have some particular construction in your head, just comment here. I don't want to do more trial and errors with this.

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,46 @@
...
+                    template, res_id.id, related_record
+                )
+                result[res_id.id]["attachments"] = invoice_attachments
+        return result
+
+    def _should_print_invoice_attachment(self):
+        if (
+            self.model_id.model == "account.move"
+            and self.report_template.report_name
+            == "l10n_ch_invoice_reports.account_move_payment_report"
```
i-vyshnevska, line 30: the report name is not mandatory here, user can replace it on any report he wants to have

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,46 @@
...
+        ):
+            return True
+
+    def generate_invoice_attachment(self, template, res_id, related_record):
+        self._render_template(template.report_name, template.model, res_id)
+        # force translation of attachment's name according to partner's lang
+        self = self.with_context(lang=related_record.partner_id.lang)
+        report_name = _("invoice_%s_with_payslip.pdf") % related_record.name.replace(
+            "/", "_"
+        )
```
i-vyshnevska:
```suggestion
        report_name = template.report_name
```
can we use template.report_name to stay generic?

l10n_ch_invoice_reports/models/account_invoice.py
i-vyshnevska: suppose it should prevent the generation of the attachment in hardcoded implementation on odoo side
https://github.com/odoo/odoo/blob/7fa5965de1c06d2f49e9a2f22a2686d99a19738a/addons/l10n_ch/models/mail_template.py#L36

the comment to method to be updated

l10n_ch_invoice_reports/models/mail_template.py
ajaniszewska-dev: we always can but afair i did like that because otherwise it wasn't correctly translated.

l10n_ch_invoice_reports/models/account_invoice.py
simahawk: The comment should say "prevent generation of attachment $here" w/ the link to or a reference to the place where it happens.
Moreover, it would be nice to propose a change to core for v >= 14 or 16 at least, to delegate that check to a new method `def _l10n_ch_isr_valid(self, related_record):`

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -0,0 +1,47 @@
...
+    def generate_email(self, res_ids, fields=None):
+        # Method overridden in order remove QR/ISR payslips and use
+        # Invoice report with payslip generated by this module
+        # https://github.com/odoo/odoo/blob/13.0/addons/l10n_ch/models/mail_template.py#L12
+
+        result = super().generate_email(res_ids, fields)
+        if self._should_print_invoice_attachment():
+            res_ids_to_templates = self.get_email_template(res_ids)
+            for res_id in self.env["account.move"].browse(res_ids):
+                related_record = self.env[self.model_id.model].browse(res_id.id)
```
simahawk:
```suggestion
            for invoice in self.env[self.model_id.model].browse(res_ids):
```

l10n_ch_invoice_reports/models/mail_template.py
simahawk: You mean that we should make this configurable?

l10n_ch_invoice_reports/models/mail_template.py
simahawk: if you keep this call to _render_template please motivate it w/ a comment

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -16,14 +16,10 @@ def generate_email(self, res_ids, fields=None):
...
         if self._should_print_invoice_attachment():
-            res_ids_to_templates = self.get_email_template(res_ids)
-            for res_id in self.env["account.move"].browse(res_ids):
-                related_record = self.env[self.model_id.model].browse(res_id.id)
-                template = res_ids_to_templates[res_id.id]
-                invoice_attachments = self.generate_invoice_attachment(
-                    template, res_id.id, related_record
-                )
-                result[res_id.id]["attachments"] = invoice_attachments
+            self.get_email_template(res_ids)
```
simahawk: is this call needed here? The result is not used at all

l10n_ch_invoice_reports/models/mail_template.py
i-vyshnevska: the issue of email generation is that implementation in odoo doesn't rely on the report attached to template, the only thing we need here is to check that report is there, which one - doesn't matter 
non-blocking thou

l10n_ch_invoice_reports/models/mail_template.py
ajaniszewska-dev: @simahawk  my mistake, we don't need it anymore. removed.

l10n_ch_invoice_reports/models/mail_template.py
simahawk: Please add a comment regarding this @i-vyshnevska's comment

l10n_ch_invoice_reports/models/mail_template.py
```diff
@@ -34,14 +29,11 @@ def _should_print_invoice_attachment(self):
...
         # force translation of attachment's name according to partner's lang
-        self = self.with_context(lang=related_record.partner_id.lang)
-        report_name = _("invoice_%s_with_payslip.pdf") % related_record.name.replace(
-            "/", "_"
-        )
+        self = self.with_context(lang=invoice.partner_id.lang)
+        report_name = _("invoice_%s_with_payslip.pdf") % invoice.name.replace("/", "_")
         report_xml_id = self.report_template.xml_id
-        report_pdf = self.env.ref(report_xml_id).render_qweb_pdf([res_id])[0]
+        report_pdf = self.env.ref(report_xml_id).render_qweb_pdf([invoice.id])[0]
```
simahawk: Am I wrong or `self.report_template` IS already the `ir.actions.report` you need? 
If that's true, you don't need to use its xid to find it again.

simahawk (commented):
2 more comments. Once done, please:
1. squash in 1 commit
2. remove "13.0" from the commit msg
3. keep an extended explanation of what has been added and why in the commit description

Thanks!

simahawk (approved):

PR #662, opened 2022-07-22
[MIG][l10n_ch_base_bank] backport of features from v. 12.0 to 11.0

patrickt-oforce:
Backport of features from v 12 to v 11, adding pre-migration to remove view, because a field was removed

**Don't bump when merge**

### DIFF
--- a/l10n_ch_base_bank/models/invoice.py
+++ b/l10n_ch_base_bank/models/invoice.py
@@ -1,6 +1,8 @@
 # Copyright 2012-2019 Camptocamp
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
-from odoo import models, api, _
+import re
+
+from odoo import _, api, models
 from odoo.tools import mod10r
 from odoo import exceptions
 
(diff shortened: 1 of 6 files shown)


l10n_ch_base_bank/models/invoice.py
```diff
@@ -57,71 +59,71 @@ def _get_reference_type(self):
...
+                    len(self.reference) == 15):
+                return True
+            if mod10r(self.reference[:-1]) != self.reference:
+                raise exceptions.ValidationError(
+                    _('Invalid ISR Number (wrong checksum).')
+                )
+        elif self.reference_type == 'qrr':
+            if not self.reference:
+                return False
+            if re.match(r'^(\d{27}|\d{2}( \d{5}){5})$', self.reference):
```
dcoopennet, line 115: I think there might be an issue with this regular expression. Can you explain me what is the purpose of the expression ?

patrickt-oforce: Hi @dcoopennet, the regular expression is got from the v12 of the module, but the reg ex try to match a valid QRR Iban code.
The regular expression is the same as here: https://github.com/OCA/l10n-switzerland/blob/12.0/l10n_ch_base_bank/models/invoice.py#L93
Because we need to mantain capability of use ISR Code or QRR i need to integrate QRR parts from the modules of successive version

PR #665, opened 2022-07-22
[MIG][l10n_ch_payment_slip] Backport from v. 12.0

patrickt-oforce:

PR #663, opened 2022-07-22
[MIG][l10n_ch_fix_isr_reference] Backport from v. 12.0

patrickt-oforce:

PR #666, opened 2022-07-22
[MIG][l10n_ch_qriban] Backport from v. 12.0

patrickt-oforce:


patrickt-oforce:
@OCA/local-switzerland-maintainers this depends on #662 as for other modules, can #662 #663 #665 and this be merged into a single one?

PR #675, opened 2022-09-28
[14.0][ADD] ebill_postfinance

TDu:


ebill_postfinance/__manifest__.py
```diff
@@ -0,0 +1,38 @@
...
+    "website": "https://github.com/OCA/l10n-switzerland",
+    "depends": [
+        "account",
+        "account_invoice_export",
+        "base_ebill_payment_contract",
+        "l10n_ch_base_bank",
+        "l10n_ch_qriban",
+        "queue_job",
+        "sale",
+        "sale_stock",
```
santostelmo: Why we need `sale_stock` to be installed ? This will force the stock module to be installed in the system while some companies don't use stock module at all

TDu: It is installed because the module adds  information about the delivery note https://github.com/OCA/l10n-switzerland/pull/675/files#diff-a2170ebdcc7a37e4c7e59582666f475e8805d9059f2798b1cbb04e951f50f549R47 in the xml generated.

ebill_postfinance/models/ebill_postfinance_service.py
```diff
@@ -0,0 +1,153 @@
+# Copyright 2022 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+import logging
+import logging.config
+
+# from ..ebilling_postfinance.ebilling_postfinance import ebilling_postfinance
```
yvaucher: To remove?

ebill_postfinance/models/ebill_postfinance_service.py
```diff
@@ -0,0 +1,153 @@
...
+        res = service.upload_files(transaction_id, file_type, data)
+        return res
+
+    def get_invoice_list(self, archive_data=False):
+        service = self._get_service()
+        res = service.get_invoice_list(archive_data)
+        return res
+
+    def get_process_protocol_list(self, archive_data=False):
+        # Is this the processing result of an invoice ?
```
yvaucher, line 128: Should it translate as a TODO ?

ebill_postfinance/models/ebill_postfinance_service.py
```diff
@@ -0,0 +1,153 @@
...
+    def get_ebill_recipient_subscription_status(self, recipient_id):
+        service = self._get_service()
+        res = service.get_ebill_recipient_subscription_status(recipient_id)
+        return res
+
+    def get_registration_protocol_list(self, archive_data=False):
+        service = self._get_service()
+        res = service.get_registration_protocol_list(archive_data)
+        for registration_protocol in res or {}:
+            self.get_registration_protocol(registration_protocol.CreateDate)
```
yvaucher, line 142: the value returned from this function is not used

ebill_postfinance/models/ebill_postfinance_service.py
```diff
@@ -0,0 +1,153 @@
...
+        string="Use Paynet/SIX format",
+        help="Enable use of legacy SIX/Paynet invoice format.",
+    )
+    operation_timeout = fields.Integer(
+        string="HTTP Timeout",
+        default="600",
+        help="Timeout for each HTTP (GET, POST) request in seconds.",
+    )
+
+    def _get_service(self):
```
yvaucher, line 61: Does it create a new session each time, can't we reuse a session by storing the current service?

yvaucher (changes requested):

TDu:
@yvaucher  Thank you for the review, I will attend them.
For now I extracted the stock related code into `ebill_postfinance_stock` see last commit 17b06e2ea2198e16cfbf9e258124206f06dfc838

ebill_postfinance/__manifest__.py
TDu: Split into a new `ebill_postfinance_stock` module the base module should not have any stock related dependencies.

ebill_postfinance/models/ebill_postfinance_service.py
TDu: Questioning the use of this call to the api.
The situation at the moment is, all possible call to the api are implemented, so they can be used and tested.
But they are not all used yet by the module for real.
There is the roadmap file that contains some more information.

ebill_postfinance/models/ebill_postfinance_service.py
TDu: This call is related to the contract subscription update through the web service. And not yet implemented, actually when I tested it I could never get any data.
I learn today that the subscription update through the web service is only implemented for b2ebanking or b2c.

ebill_postfinance/models/ebill_postfinance_service.py
TDu: This is a very good point. And yes I believe there is some improvement to be done on this...

ebill_postfinance/models/ebill_postfinance_service.py
TDu: done

TDu:
Previous fixup fixes the number of references included for each invoice line in the xml payload.
Adds a payload size in the data, can be useful, knowing that the web service will refuse more than 5MB.

ebill_postfinance/models/account_move.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        if self.transmit_method_id == postfinance_method:
+            contract = self.partner_id.get_active_contract(self.transmit_method_id)
+            if contract:
+                self.partner_bank_id = contract.postfinance_service_id.partner_bank_id
+
+    def _export_invoice(self):
+        """Export invoice with the help of account_invoice_export module."""
+        postfinance_method = self.env.ref(
+            "ebill_postfinance.postfinance_transmit_method"
```
i-vyshnevska, line 35:
```suggestion
            "ebill_postfinance.postfinance_transmit_method", raise_if_not_found=False
```

ebill_postfinance/models/account_move.py
```diff
@@ -0,0 +1,139 @@
...
+
+    _inherit = "account.move"
+
+    @api.onchange("partner_id", "company_id")
+    def _transmit_method_partner_change(self):
+        super()._transmit_method_partner_change()
+        if self.move_type not in ("out_invoice", "out_refund"):
+            return
+        postfinance_method = self.env.ref(
+            "ebill_postfinance.postfinance_transmit_method"
```
i-vyshnevska, line 25:
```suggestion
            "ebill_postfinance.postfinance_transmit_method", raise_if_not_found=False
```

ebill_postfinance/models/account_move.py
TDu: If this record is not present, nothing will work on the module. So better to have the error raised IMO
Same for the comment below.

ebill_postfinance/models/account_move.py
TDu: If this record is not present, nothing will work on the module. So better to have the error raised IMO

ebill_postfinance/views/ebill_postfinance_service.xml
```diff
@@ -0,0 +1,136 @@
...
+                            />
+                            <field
+                                name="partner_bank_id"
+                                options="{'no_create': True}"
+                            />
+                        </group>
+                        <group>
+                            <field name="biller_id" />
+                            <field name="username" />
+                            <field name="password" />
```
santostelmo:
```suggestion
                            <field name="password"  password="True"/>
```
Could you please hide the password value?

ebill_postfinance/models/ebill_postfinance_invoice_message.py
```diff
@@ -94,13 +94,13 @@ def _compute_payload_size(self):
...
-                [
-                    fields.Datetime.now().strftime("%Y%m%d%H%M"),
-                    record.invoice_id.name.replace("/", "").replace("_", ""),
-                ]
-            )
+        self.ensure_one()
+        self.transaction_id = "-".join(
+            [
+                fields.Datetime.now().strftime("%y%m%d%H%M%S"),
+                self.invoice_id.name.replace("/", "").replace("_", ""),
```
simahawk, line 106: I would use slugify to make sure there's no weird char.

simahawk:
@TDu what's the overall state here?

ebill_postfinance/views/ebill_postfinance_service.xml
TDu: done

TDu:
I squashed all previous commit and then the last fixup adds the credit note to be sent to the server.

sebalix:
@TDu so we are good to merge this?

ecino:
Maybe you could increase the coverage of the tests? Also you can force-push to retrigger the runboat. Is this ready to be merged otherwise?

TDu:
@sebalix  The required Python library needed to be on Pypi. This is done.
@ecino  More tests would be nice, but I will improve them when migrating to 16.0
@simahawk  IMO this is ready for merge
@yvaucher Could you update your review

ebill_postfinance/models/account_move.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+            activity = self.activity_schedule(
+                activity_type, summary="Invoice rejected by Postfinance", note=message
+            )
+        # error_log = values.get("error_detail")
+        # if not error_log:
+        #     error_log = _("An error of type {} occured.").format(
+        #         values.get("error_type")
+        #     )
+        # activity.note += "<div class='mt16'><p>{}</p></div>".format(error_log)
```
ecino: This can be removed if it's not needed.

ecino (commented):
Should you also add `server-env` and `edi` in file `oca_requirements.txt` ?

ebill_postfinance/models/ebill_postfinance_service.py
ecino: Maybe decorating the method with `@ormcache` would work?

ebill_postfinance/models/account_move.py
TDu: Removed.

TDu:
> Should you also add `server-env` and `edi` in file `oca_requirements.txt` ?

With the introduction of the use of PyPi to find depending modules, this is not needed anymore, as far as I know, in fact Travis runs without errors.
Thanks for your review.

sebalix (approved):

yvaucher (approved):
My remarks for improvements can be addressed later, better to have a version we can migrate to other version than none.

PR #685, opened 2022-12-09
[16.0][MIG] l10n_ch_invoice_reports: Migrate to version 16.0

sonhd91:


TDu:
@sonhd91 It needs a rebase.

yvaucher:
The ISR part of this module is obsolete and should be removed, not sure this module is still useful or not without it.

l10n_ch_invoice_reports/models/account_invoice.py
```diff
@@ -0,0 +1,51 @@
...
+    _inherit = "account.move"
+
+    # add context key dependency to prevent generation of attachment inside
+    # https://github.com/odoo/odoo/blob/15.0/addons/l10n_ch/models/mail_template.py#L12
+    @api.depends_context("invoice_report_no_attachment")
+    def _compute_l10n_ch_isr_valid(self):
+        if self.env.context.get("invoice_report_no_attachment", False):
+            self.update({"l10n_ch_isr_valid": False})
+        else:
+            return super()._compute_l10n_ch_isr_valid()
```
ajaniszewska-dev, line 18: we dont need to handle isr anymore, try to remove any isr-dependent code and leave only QR

nilshamerlinck:
Hello @yvaucher and @ajaniszewska-dev and happy new year!

Since [odoo 16.0](https://github.com/odoo/odoo/pull/82341):

> All invoices for which a print is asked will be printed with a QR bill if it is possible. If not, this will check if printing in the ISR format is possible. Lastly, if none is possible, the classic invoice will be printed.

True on portal side as well (when downloading the invoice).

So I believe this module is not needed anymore indeed. If somebody thinks otherwise, we'll re-open that PR (without ISR).

PS: for the record, it seems Odoo [intended to remove the ISR part from 16.0](https://github.com/odoo/odoo/blob/16.0/addons/l10n_ch/models/account_invoice.py#L17) but didn't somehow.

PR #682, opened 2022-11-24
l10n_ch_base_bank remove isr check

santostelmo:
ISR is not used anymore in Switzerland => We drop ` _check_bank_type_for_type_isr`

PR #659, opened 2022-07-19
[15.0][MIG] l10n_ch_base_bank: Migrate to version 15.0

sonhd91:

PR #686, opened 2023-01-04
[16.0] [MIG] `l10n_ch_base_bank`

ivantodorovich:
It follows up after:
- https://github.com/OCA/l10n-switzerland/pull/659

It includes:
- https://github.com/OCA/l10n-switzerland/pull/682

And a number of other improvements, each in a different commit to ease review


yvaucher (commented):
The ISR part is now obsolete since end of last year.

We should drop this module as it only exists for that.

ivantodorovich:
Alright, closing this PR then

PR #687, opened 2023-01-04
[16.0] [MIG] `l10n_ch_bank`

ivantodorovich:
* Re-download and re-format the bank csv.
* Remove useless columns from import.
* Update README with instructions on how to do it.

Depends on:
- https://github.com/OCA/l10n-switzerland/pull/686


yvaucher:
The purpose of this module was to provide BIC for banks. However, it was only required for ISR / ISR-B. And the list of bank might be quite outdated.

I suggest to drop that module too.

PR #690, opened 2023-01-09
[14.0][FIX] ebill_paynet: missing transmit method in tests

StephaneMangin:
As the transmit method is required by the get_active_contract from partner when account move create_paynet_message is called, it should be configured in test as well.


TDu (approved):
Not sure why it did not fail before, but LG

StephaneMangin:
> Not sure why it did not fail before, but LG

Yes it's kind of strange. I looked out over the changes since the last valid commit and found... nothing.

SilvioC2C (approved):
LGTM

PR #691, opened 2023-01-20
[ADD] l10n_ch_cash_rounding

ajaniszewska-dev:


l10n_ch_cash_rounding/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    invoice_cash_rounding_id = fields.Many2one(
+        "account.cash.rounding",
+        readonly=False,
+        compute="_compute_invoice_cash_rounding",
+        states={},
```
grindtildeath:
```suggestion
        compute="_compute_invoice_cash_rounding",
        states={'draft': [('readonly', False)]},,
```
This will allow to define the cash rounding only if the move is not posted yet.

l10n_ch_cash_rounding/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+    def _compute_invoice_cash_rounding(self):
+        for move in self:
+            if move.move_type in [
+                "out_invoice",
+                "out_refund",
+            ]:
+                if move.currency_id.id == self.env.ref("base.CHF").id:
+                    move.invoice_cash_rounding_id = self.env.ref(
+                        "l10n_ch_cash_rounding.swiss_cash_rounding"
+                    ).id
```
grindtildeath:
```suggestion
        swiss_cash_rounding = self.env.ref(
            "l10n_ch_cash_rounding.swiss_cash_rounding",
            raise_if_not_found=False
        )
        if not swiss_cash_rounding:
            return
        for move in self:
            if move.move_type in [
                "out_invoice",
                "out_refund",
            ]:
                if move.currency_id == self.env.ref("base.CHF"):
                    move.invoice_cash_rounding_id = swiss_cash_rounding
```
This will ensure the function won't break if the cash rounding record is deleted.

grindtildeath (commented):
IMO it would still be good to make this somehow configurable through the UI.

Rethinking about it, there used to be an OCA module before the cash rounding model did exist in Odoo where IIRC the rounding was applied automatically: https://github.com/OCA/account-invoicing/tree/10.0/account_invoice_rounding

Since the cash rounding model didn't exist back in this time, maybe we can do something like creating a module account_cash_rounding_auto in OCA/account-invoicing. This module would:
  - Redefine the field as compute and provide the compute function.
  - Add a O2m field `auto_apply_currency_ids` on account.cash.rounding .
  - Add a M2o field `auto_cash_rounding_id` on `res.currency`
  - In the compute function, it would apply the cash rounding of the currency.

This way, it wouldn't have to be specific for switzerland or CHF (as other countries such as sweden and probably others also need a cash rounding).

Would also be great to have some tests. :grimacing: 

EDIT: I also think this would make sense to avoid the dependency on l10n_ch since IMO it's more related to the currency than the chart of account one is using.

l10n_ch_cash_rounding/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import api, fields, models
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    invoice_cash_rounding_id = fields.Many2one(
+        "account.cash.rounding",
+        compute="_compute_invoice_cash_rounding",
+        states={"draft": [("readonly", False)]},
```
i-vyshnevska, line 13:
```suggestion
        states={"draft": [("readonly", False)]},
        store=True,
```

l10n_ch_cash_rounding/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+
+{
+    "name": "Switzerland - CHF Cash Rounding",
+    "summary": "Apply cash rounding for CHF currency",
+    "version": "15.0.1.0.0",
+    "author": "Camptocamp,Odoo Community Association (OCA)",
+    "category": "Localization",
+    "website": "https://github.com/OCA/l10n-switzerland",
+    "license": "AGPL-3",
+    "depends": ["account", "l10n_ch"],
```
i-vyshnevska, line 13: actually, it doesn't depend on l10_ch

i-vyshnevska (approved):

l10n_ch_cash_rounding/models/account_move.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @api.depends("move_type", "currency_id")
+    def _compute_invoice_cash_rounding(self):
+        swiss_cash_rounding = self.env.ref(
+            "l10n_ch_cash_rounding.swiss_cash_rounding", raise_if_not_found=False
+        )
+        if not swiss_cash_rounding:
+            return
+        for move in self:
+            if move.move_type in [
```
TDu, line 24: This function https://github.com/odoo/odoo/blob/5e98953d67042dcffa978ab7d240d8677e847456/addons/account/models/account_move.py#L2556 can be used for that.

TDu (commented):
It needs a small unit test

PR #697, opened 2023-04-28
[IMP] l10n_invoice_reports QR and invoice attachments merged

santostelmo:
### Context
Currently QR and invoice report attachments are generated separately. We want single generate a single attachment and store it in the DB


cyrilmanuel (approved):
LG

PR #708, opened 2023-07-18
[14.0][FIX] Disable ebill_paynet unit test

TDu:
This module should not be used anymore because the service has been shut down.
Use ebill_postfinance instead (still in PR form)


sebalix (approved):

sebalix:
Link to `ebill_postfinance`: https://github.com/OCA/l10n-switzerland/pull/675

PR #709, opened 2023-08-08
[16.0] [MIG] l10n_ch_mis_reports: Migration to 16.0

jguenat:
Migration with following refactor
-  Use formulas based on account code instead of account tag, [l10n_ch_account_tags](https://github.com/OCA/l10n-switzerland/tree/14.0/l10n_ch_account_tags) not required anymore.  It was not very practical to have to assign a tag to each account.
- move data to xml noupdate records : you can modify your templates/styles without loosing the modifications on update.
- simplify styles

reviews/opinions are welcome :)


ecino (commented):
I have no experience with this module in the previous versions, but it looks good to me (only code review). One quetion, shouldn't we put the XML files in English and provide translations instead of directly entering records in French?

jguenat:
> I have no experience with this module in the previous versions, but it looks good to me (only code review). One quetion, shouldn't we put the XML files in English and provide translations instead of directly entering records in French?

Thanks for your review !

Yes it would be better but I don't know the english translation.  As this is specific for switzerland I guess it's not a big deal if we put it in french, people can then translate it with weblate if needed.

ecino (approved):

l10n_ch_mis_reports/data/mis_report_kpi_pl.xml
```diff
@@ -0,0 +1,245 @@
+<odoo noupdate="1">
+
+    <record id="mis_report_kpi_0001" model="mis.report.kpi">
+        <field name="name">CH_30</field>
+        <field name="description">Chiffre d'affaires</field>
+        <field
+            name="expression"
+        >-balp[('code', '>=', '3'),('code', '&lt;', '39')][]</field>
+        <field name="style_id" ref="style_details" />
```
eicher31: Thanks for your work.
Could you juste change the style here with
```suggestion
        <field name="style_id" ref="style_header" />
```
Otherwise it's all ok for me.

jguenat: Thanks for your review. Indeed it's nicer with the header style.

eicher31 (approved):
Ok for me

eicher31:
@ecino could you merge it? Thanks

PR #710, opened 2023-09-05
[15.0][FIX] l10n_ch_invoice_reports: remove controller

janikvonrotz:
I got a 500er in the portal view caused by the controller. It seems to be 1:1 ported from Odoo 14.0. The API has changed and therefore I propose to remove it entirely.


ecino:
@sonhd91 Can you review?

TDu:
Removing the whole controller because there could be a bug seems a bit harsh...
Also I tested on runboat from the 15.0 branch an I did not have any issue :thinking: 
The functionality actually works (using the Preview button on the invoice)

PR #711, opened 2023-10-05
[14.0] [ADD] l10n_ch_credit_control_invoices

santostelmo:
### Context
Credit account_credit_control module module helps to identify outstanding debt beyond tolerance level and setup followup method. It allows you to generate s summary report (per partner) with a reference of overdue invoices.

With that module we have the option to merge into credit control summary report, all the invoices (eventually with QR code) to be paid by the customer.


Ricardoalso (approved):
Not an expert, but code LGTM

l10n_ch_credit_control_invoices/wizard/credit_control_printer.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2023 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import _, models
+from odoo.exceptions import UserError
+
+
+class CreditControlPrinter(models.TransientModel):
+    _inherit = "credit.control.printer"
+    """Print lines"""
```
Ricardoalso: Probably forgotten comment :thinking:

santostelmo: yes thank you. 
I cleaned that.

TDu:
Looking at this functionality I see nothing specific to the Swiss localization, it could be helpful for other localization.
IMO this should be proposed on https://github.com/OCA/credit-control
Also, it needs some test coverage.

santostelmo:
> Looking at this functionality I see nothing specific to the Swiss localization, it could be helpful for other localization. IMO this should be proposed on https://github.com/OCA/credit-control Also, it needs some test coverage.

It depends of `l10n_ch_invoice_reports` as we want to print invoices + qr bill. I wanted to take advantage to that module that allows you to merge invoice + qr bill

But I check if I can avoid the dependency merging the 3 reports in memory (credit control summary + invoices + qr bills)

PR #723, opened 2024-05-31
[14.0][FIX] Enable tests to be run with Github actions

TDu:
Unit tests are not running on the 14.0 branch. This is due to Travis that has been phased out.
Running a `copier update` to enable the use of Github actions :crossed_fingers:


TDu:
@yvaucher Quick review, please

PR #727, opened 2024-06-12
16.0 [MIG] l10n_ch_pain_base module (without l10n_ch_base_bank)

davidwul:
superseed #722 
@BT-anieto 
here is the l10n_ch_pain_base module only, is that what you meant? I'm not proficiency with github


BT-anieto:
@davidwul You have to follow https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-16.0

PR #735, opened 2024-09-11
[14.0][FEAT]: Improve l10n_ch_invoice_reports

victorvermot:
We want the customer to be able to chose which reports will be printed when choosing to print Invoice with payment slip(s)

I added some settings to be able to chose which reports are printed.
fmel_722_1

### Result 
- You can now print reports that show payment with the option "print invoice with payment slip(s)"
fmel_722_3


cyrilmanuel (approved):

Ricardoalso (approved):
Work LGTM; some additional unitests would be great

ecino (approved):

PR #737, opened 2024-09-18
[16.0] Add module l10n_ch_credit_control_dunning_fees_qr_report

grindtildeath:
This module will ensure dunning fees are considered when generating swiss
QR-bill from credit control lines.


l10n_ch_credit_control_dunning_fees_qr_report/models/credit_control_communication_report.py
```diff
@@ -0,0 +1,108 @@
...
+        self = self.with_context(credit_control_run_print_report=True)
+        res = super()._get_report_values(docids, data=data)
+        dunning_fees_qr_amounts = self.env.context.get("qr_dunning_fees_amounts", {})
+        if not dunning_fees_qr_amounts:
+            return res
+        qr_code_urls = {}
+        invoices = self.env["account.move"].browse(docids)
+        for invoice in invoices:
+            if invoice.id not in dunning_fees_qr_amounts:
+                continue
```
simahawk, line 24:
```suggestion
        invoices = self.env["account.move"].browse(list(dunning_fees_qr_amounts.keys()))
        for invoice in invoices:
```
You know already the ones to browse no?

l10n_ch_credit_control_dunning_fees_qr_report/models/account_move.py
```diff
@@ -0,0 +1,17 @@
...
+    _inherit = "account.move"
+
+    def get_qr_amount(self):
+        qr_dunning_fees_amounts = self.env.context.get("qr_dunning_fees_amounts", {})
+        if self.id in qr_dunning_fees_amounts:
+            return "{:,.2f}".format(qr_dunning_fees_amounts[self.id]).replace(
+                ",", "\xa0"
+            )
+        else:
+            return "{:,.2f}".format(self.amount_residual).replace(",", "\xa0")
```
simahawk, line 17:
```suggestion
        if self.id in qr_dunning_fees_amounts:
            return "{:,.2f}".format(qr_dunning_fees_amounts[self.id]).replace(
                ",", "\xa0"
            )
        return "{:,.2f}".format(self.amount_residual).replace(",", "\xa0")
```
any comment to explain why this is needed?

l10n_ch_credit_control_dunning_fees_qr_report/models/credit_control_communication_report.py
```diff
@@ -0,0 +1,108 @@
...
+                invoice.currency_id,
+                invoice.partner_id,
+                qr_method="ch_qr",
+                silent_errors=False,
+            )
+        res["qr_code_urls"].update(qr_code_urls)
+        return res
+
+
+class Report(models.Model):
```
simahawk: this should go to its own file

l10n_ch_credit_control_dunning_fees_qr_report/models/credit_control_communication_report.py
```diff
@@ -0,0 +1,108 @@
...
+            "account_credit_control.report_credit_control_summary"
+        )
+        report = self._get_report(report_ref)
+        report_name = report.report_name
+        if report_name != credit_control_communication_report_name:
+            return super()._render_qweb_pdf_prepare_streams(
+                report_ref, data, res_ids=res_ids
+            )
+
+        collected_streams = OrderedDict()
```
simahawk: maybe move to a separate method like  `_pdf_prepare_streams_cc_dunning_fees`?

simahawk (approved):
Non blocking remarks. Pre-approving.
Any chance to have tests?

grindtildeath:
@simahawk Out of moving class to its dedicated file, I would prefer keeping code as it to avoid any unforeseen issue since it's been working in production for a while.

Also I'm not the original author, most of the code was from @ajaniszewska-dev , so I cannot necessarily answer all these questions.

A test would be good yes.

PR #740, opened 2024-10-15
[18.0][MIG] l10n_ch_adr_report: Migration to 18.0

xaviedoanhduy:
### Depends on:

- https://github.com/OCA/community-data-files/pull/211

### Note

- Qty done and Reserved qty are merged into a single column in commit: https://github.com/odoo/odoo/commit/7dda6bb92715ea25b2818a62fec5e646f3678b81

### Result
[Swiss Dangerous Goods Report.pdf](https://github.com/user-attachments/files/17291330/Swiss.Dangerous.Goods.Report.pdf)


l10n_ch_adr_report/report/dangerous_delivery_report.py
```diff
@@ -116,19 +113,10 @@ def _get_DG_move_line_vals(self, moves):
         # unit measurement on stock is not considered
         result = []
         for product, move_lines in moves:
-            qty = 0
-            for rec in move_lines:
-                if rec.state == "done":
-                    qty += rec.quantity_done
-                else:
-                    qty += rec.product_uom_qty
+            qty = sum(line.quantity for line in move_lines)
```
TDu: I am not sure this is correct, because with the change in Odoo core we still have the `product_uom_qty` field.
I think this
```
qty = 0
for rec in move_lines:
    if rec.picked:
        qty += rec.quantity
    else:
        qty += rec.product_uom_qty
```
would be better

TDu:
Also adding some small unit tests would be a good improvement for this migration

l10n_ch_adr_report/report/dangerous_delivery_report.py
xaviedoanhduy: thanks @TDu, i see that the use of `picked` field is not quite like [its definition](https://github.com/odoo/odoo/blob/5bb1f22984506188f5300ae765a70b813f32dcad/addons/stock/models/stock_move.py#L125) in this case. but maybe i was a bit confused about `product_uom_qty` so i will restore the original code and just edit `quantity` to replace the outdated `quantity_done`

TDu (approved):

sebalix (approved):

sebalix:
Hello @xaviedoanhduy , can you rebase your PR now that the dependency is merged? Thank you :pray:

xaviedoanhduy:
> Hello @xaviedoanhduy , can you rebase your PR now that the dependency is merged? Thank you 🙏

hi @sebalix, it has been rebase!

PR #741, opened 2024-11-12
[16.0][ADD] l10n_ch_qr_bill_without_amount

santostelmo:
Allows you to print QR invoices without a predefined amount. The QR invoice without an amount is based on the payment mode field `Qr Bill Without Amount`


cyrilmanuel (approved):

l10n_ch_qr_bill_without_amount/i18n/l10n_ch_qr_bill_without_amount copy.pot
```diff
@@ -0,0 +1,28 @@
...
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: \n"
+"Plural-Forms: \n"
+
+#. module: l10n_ch_qr_bill_without_amount
+#: model:ir.model.fields,field_description:l10n_ch_qr_bill_without_amount.field_account_bank_statement_line__qr_bill_amount
+#: model:ir.model.fields,field_description:l10n_ch_qr_bill_without_amount.field_account_move__qr_bill_amount
+#: model:ir.model.fields,field_description:l10n_ch_qr_bill_without_amount.field_account_payment__qr_bill_amount
+msgid "Qr Bill Amount"
+msgstr "Montant de la facture Qr"
```
Ricardoalso: Translation done in the wrong file

l10n_ch_qr_bill_without_amount/views/payment_mode.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="account_payment_mode_form" model="ir.ui.view">
+        <field name="name">account.payment.mode.form</field>
+        <field name="model">account.payment.mode</field>
+        <field name="inherit_id" ref="account_payment_mode.account_payment_mode_form" />
```
Ricardoalso, line 6: account_payment_mode module isn't in the manifest file

santostelmo: dependency account_payment_mode already in account_payment_partner

Ricardoalso (approved):

l10n_ch_qr_bill_without_amount/models/account_move.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class AccountMove(models.Model):
+    _inherit = "account.move"
+
+    qr_bill_amount = fields.Char(
+        compute="_compute_qr_bill_amount",
+    )
+
+    @api.depends("amount_residual")
```
TDu, line 14:
```suggestion
    @api.depends("amount_residual", "payment_mode_id.qr_bill_without_amount")
```

TDu (commented):
There is already a very similar module in 13.0 and 17.0
https://github.com/OCA/l10n-switzerland/tree/17.0/l10n_ch_qr_no_amount
It does not have the possible configuration with the payment mode, but it would make sense to merge them  in the future IMO

l10n_ch_qr_bill_without_amount/models/account_move.py
santostelmo: Thank you @TDu , what a pity I didn't find the model in version 17 before. It makes more sense to backport this module and create a glue module.

santostelmo: @TDu I just realized that the 17.0 module is not working as expected.
When we scan th QR we still see the amount:
image
You can try it here : https://qr-rechnung.net/#/scan

santostelmo: Another thing that is not possible is to print invoice **with** and **without** amount in batch.

TDu: Ok, can we keep in the readme or roadmap of this module that the other one exists and we should try to merge them or regroup their common functionality in some way...
Also I think there still is a missing field in the api.depends (see comment above)

grindtildeath: > When we scan th QR we still see the amount

That's a bug of the module... (EDIT: Fixed in #745 )

> Another thing that is not possible is to print invoice with and without amount in batch

Feel free to provide an improvement on the existing module, there's no reason to have two different modules for that

grindtildeath (changes requested):
As written in my comment above, the same module already exists, so we should close this and eventually improve the existing module.

santostelmo:
> As written in my comment above, the same module already exists, so we should close this and eventually improve the existing module.

It is not the same module. How can I answer the specific need of the client to send invoices with and without amount by email ? That is not possible in the module [l10n_ch_qr_no_amount](https://github.com/OCA/l10n-switzerland/tree/17.0/l10n_ch_qr_no_amount). Probably if you see a way to merge that feature in `l10n_ch_qr_no_amount` we could refactor that in the future.

grindtildeath:
@santostelmo The module provides the report without amount. Choosing which report has to be sent by e-mail is a matter of integrating it for your customer.

santostelmo:
> @santostelmo The module provides the report without amount. Choosing which report has to be sent by e-mail is a matter of integrating it for your customer.

Customer must be able to mix both reports. For example:
I select a set of invoices in tree view. I print report and I get a single file with invoices with and without amount.
I don't know how to achieve that with [l10n_ch_qr_no_amount](https://github.com/OCA/l10n-switzerland/tree/17.0/l10n_ch_qr_no_amount).. 

I'm thinking to convert that dev in customer specific.

grindtildeath:
@santostelmo To me, what you describe is a specific requirement.

You can handle that in a specific module that depends on `l10n_ch_qr_no_amount`, in order to keep the customization of the report (ie no amount encoded, hash marks displayed instead of empty amount on the slip) in the OCA.

santostelmo:
Thanks @grindtildeath . I close this PR then.

Quick note for the related project: This module should be moved to specific.

santostelmo:
Thanks @grindtildeath . I close this PR.

Quick note for related project. This module should be moved to specific.

PR #742, opened 2024-11-18
[16.0][MIG] l10n_ch_qr_no_amount

santostelmo:
Backport l10n_ch_qr_no_amount from 17.0


l10n_ch_qr_no_amount/report/swissqr_no_amount_report.py
```diff
@@ -1,7 +1,13 @@
...
 
 
 class ReportSwissNoAmountQR(models.AbstractModel):
     _name = "report.l10n_ch_qr_no_amount.qr_no_amount_report_main"
     _inherit = "report.l10n_ch.qr_report_main"
     _description = "QR-No-Amount-bill"
+
+    @api.model
+    def _get_report_values(self, docids, data=None):
+        if self._name == "report.l10n_ch_qr_no_amount.qr_no_amount_report_main":
```
grindtildeath, line 11: This condition will always evaluate to True as it's the `_name` of the `class` where the override is done.
You can cherry-pick from https://github.com/OCA/l10n-switzerland/pull/745 to solve the amount encoded in QR issue that was introduced in the migration to 17.0

PR #743, opened 2024-11-27
[18.0][MIG] l10n_ch_account_tags: Migration to 18.0

Kimkhoi3010:
Since `account.template` is no longer supported from version 17.0, the process of linking `account.tag` to `account.account` will be handled using `post_init`.


henrybackman (approved):

TDu (approved):
Thanks for the migration, that migration script makes sense without the templates.

ecino (approved):
LGTM

PR #751, opened 2025-02-11
[18.0][ADD] l10n_ch_merge_states: New module

jguenat:
Hello,
I do not really like how switzerland country states have been managed by Odoo since V16 : no translation -> 3 states per canton.
This is not great specially if you have a webshop.
This module tries to fix this in an opinionated way by choosing one language (or 4th option : keep canton main language) and archiving the others.
Many2one and Many2many fields are also reassigned based on the language choice.

Feedback welcome :)


jguenat:
@divad1196 
FYI what I was talking to you about

PR #725, opened 2024-06-11
[16.0][MIG] l10n_ch_pain_base: Migration to 16.0

BT-anieto:
Removed dependency on l10n_ch_base_bank

PR #754, opened 2025-02-18
[16.0][MIG] l10n_ch_pain_base

ecino:
Continues on #725


thomaspaulb:
@ecino Would you still have time to rebase this so that the merge commit goes away?

thomaspaulb (approved):

jguenat (approved):
LGTM
I did some testing with https://validation.iso-payments.ch/sps/validation
there is some warnings but no error

Image

Regarding test failing I thinks it's due to test data from ebill modules not compatible with bank-payment sepa modules.

thomaspaulb:
The validation against "02" format is good, which is the same as the 14.0 module handles, so from that aspect I think it can be merged. However the "03" format is out and the "02" format will soon not be accepted, so from that aspect, more needs to be done, but maybe not in the scope of this PR.

3 feb. 2026 19:44:14 Julien Guenat ***@***.***>:

ecino:
Thanks for your inputs. I agree with Thomas that the format acceptance is out of scope for now. I will need to check if we can do something about the tests before merging.

thomaspaulb:
The quickest way is to rerun copier and then in the question/answer dialogue define that one of the two conflicting modules is a rebel module and needs its own CI

4 feb. 2026 09:46:58 ecino ***@***.***>:

jguenat:
Hello,
For the tests I proposed a simple PR to fix the test data causing the issue #778 
Concerning the format acceptance we have till november 2026 to implement a new pain format.
https://www.six-group.com/en/products-services/banking-services/payment-standardization/standards/iso-20022.html#sps-2021
I took a look at official spec it seems the new swiss credit transfer pain format is based on EU pain.001.001.09
Main issue seems to be the address handling which now needs to separate street address and number.

thomaspaulb:
@jguenat About that format, some information:

- I can indeed confirm that the new Swiss format is is based on `PAIN.001.001.09`
- So much so, that in fact the one Swiss bank that we are dealing with, also accepts `PAIN.001.001.09` which makes me wonder if it is even necessary to support the Swiss format at all

To say that `PAIN.001.001.09` requires to separate street name and street number is one way of putting it, the way I prefer putting it is that in `PAIN.001.001.09` you have three choices:

- Minimal: Only put Country and City and nothing else, because nothing else is required according to the spec.
- Hybrid: Put Country, City, Zipcode and 2 AdrLine's where you don't have to separate street and street number
- Structured: Put Country, City, Street, Streetnumber and no Adrlines

We made a PR https://github.com/OCA/bank-payment/pull/1550 that supports "Minimal" and "Hybrid" and the "Minimal" is the one that is accepted by the Swiss bank

When it gets merged we will forward port to 16 and 18

If the Swiss format is truly needed then it's easy to apply to concepts of it to a PR here

jguenat:
Thanks @thomaspaulb  for the detailed information.
I think we'll still need to make some adjustment for the swiss implementation to support QRR (structured reference for special IBAN). I'll test this when your PR gets merged (not sure I can help with yours I'm quite new with pain format).

ecino:
We have green lights now! I added tests for good coverage and found a few issues while doing this (module dependency missing and no pain favor addition). For the test to work I was obliged to add the pain versions inside this module. They were on the respective credit_transfer and direct_debit submodules, do you think it would be ok to move those here? May I ask you to review once more? @thomaspaulb @jguenat

jguenat:
@ecino 
It's kind of strange to add pain flavor and have the test here because without the other module l10n_ch_pain_credit_transfer pain flavor will not work.

I'm more in favor of having bad test coverage here, it's not mandatory for merging right?

ecino:
I think you are right. Maybe I can adapt minimal testing without having to move those pain versions here which doesn't make sense in a business perspective.

ecino:
@jguenat But it's still weird that we have a portion of the code inside the module that is referencing those pain versions. Maybe we should consider refactoring that as well (https://github.com/OCA/l10n-switzerland/pull/754/changes#diff-efa70e9e1ab3cb58202c08b4defb4a066249949105d5c821a6dd59066dcb4ccfR14)

jguenat:
@ecino yes I think a big refactor is needed, also as @thomaspaulb said this pain version is soon not supported anymore (november 2026). 
I am trying to implement the new pain.001.001.09.ch.03 in version 14.0. Not quite there yet but I think I will be able to remove a lot of code inside l10n_ch_pain_base.
Also I'm not familiar with the old module l10n_ch_pain_direct_debit (not migrated since 10.0) but I think the module partition was because of him. Is it still useful?

My opinion is if we merge this now better to stick as close as possible to the old architecture despite red test coverage.

We could also wait a bit and directly implement pain.001.001.09.ch.03 in version 16.0 & 18.0

thomaspaulb:
@ecino We have the same thing in the `bank-payment` modules: [pain_flavor](https://github.com/OCA/bank-payment/blob/95f720ae3f565923cc2b348395f9424c6bc876bb/account_banking_pain_base/models/account_payment_order.py#L328) being a conditional in some of the shared functions that generate the blocks.

I think this came to be because the structure of some of these "common" blocks, for example "Address info", are shared between more than one module eg credit but also debit module, and that is why they are in the shared "base" module. To me that seems logical and does not necessarily need refactoring.

@jguenat It should be relatively easy to implement the new Swiss PAIN version in 14.0 because Swiss 02 --> 03 it's 90% the same changes as the bank-payment PAIN versions 03 --> 02, so you can just sort of copy what we did in the bank-payment PR. Or, propose a refactor to our PR that allows you to call it.

TDu (approved):
Great to see some traction in this. There is always the `/README/roadmap.rst` that can be used for keeping track of needed improvements

PR #770, opened 2025-09-29
[16.0][FIX] ebill_postfinance: Use sudo to create postfinance message

grindtildeath:


TDu:
LG, can you apply the pre-commit changes

PR #776, opened 2026-01-30
[14.0][FIX] l10n_ch_pain_base: only override address block for Swiss PAIN formats

ntsirintanis:
l10n_ch_pain_base: only override address block for Swiss PAIN formats

Before: having `l10n_ch_pain_base` installed overrided address block generation even for other countries' formats

After: Swiss address block is used for swiss pain formats, other formats keep using the generic function defined in `super` 

Somewhat related: https://github.com/OCA/bank-payment/pull/1550

### DIFF
--- a/l10n_ch_pain_base/models/account_payment_order.py
+++ b/l10n_ch_pain_base/models/account_payment_order.py
@@ -10,6 +10,10 @@ from odoo.exceptions import UserError
 class AccountPaymentOrder(models.Model):
     _inherit = "account.payment.order"
 
+    @api.model
+    def _is_ch_pain_flavor(self, pain_flavor):
+        return pain_flavor in ["pain.001.001.03.ch.02", "pain.008.001.02.ch.01"]
+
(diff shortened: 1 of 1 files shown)


thomaspaulb (approved):

thomaspaulb:
@ecino Could you review and perhaps cherry-pick to the 16.0 PR

ecino (approved):

thomaspaulb:
nobump? patch, right?

ecino:
> nobump? patch, right?

Yeah, I got too fast here....

TDu:
There is only python code change, `nobump` is ok
Thanks for contributing

thomaspaulb:
But I think if the version is not bumped, the push to PyPi is not triggered, so anyone relying on PyPi for distribution will not get the update.

thomaspaulb:
Ah, not true, it will in that case publish a "pre-release" version:

https://pypi.org/project/odoo14-addon-l10n-ch-pain-base/#history

So we're good.

PR #777, opened 2026-02-03
[16.0][MIG] l10n_ch_pain_credit_transfer: Migration to 16.0

jguenat:
depend on #754


victor-champonnois:
Thanks for the PR! 
I made some tests on a swiss database. 
I made a payment order for a belgian supplier. (the order is not SEPA, I don't know if it matters). 
I had the error : 

_The generated XML file is not valid against the official XML Schema Definition. The generated XML file and the full error have been written in the server logs. Here is the error, which may give you an idea on the cause of the problem : Element '{http://www.six-interbank-clearing.com/de/pain.001.001.03.ch.02.xsd}Othr': This element is not expected. Expected is one of ( {http://www.six-interbank-clearing.com/de/pain.001.001.03.ch.02.xsd}BIC, {http://www.six-interbank-clearing.com/de/pain.001.001.03.ch.02.xsd}ClrSysMmbId )., line 41_

Here is the xml file
[test.xml](https://github.com/user-attachments/files/27131101/xml.version.1.xml)

ecino:
@jguenat Can you rebase and make sure the tests are passing?

jguenat:
@ecino rebased
concerning @victor-champonnois issue I think the best way would be to implement the new pain.001.001.09.ch.03 that I started here #779 . This new swiss implementation is (at first view) simpler and closer to SEPA european standards with should work better in your case.
I'll look into it in the near future.

ecino (approved):
Looks good to me, thanks for the work!

PR #780, opened 2026-03-02
[18.0][MIG] l10n_ch_mis_reports: Migration to 18.0

jguenat:
Hello,

I take the opportunity of this migration to propose an improvement to the balance sheet (in a separate commit).
The idea is to display unaffected earnings on the balance sheet to better reflect Odoo [official documentation of closing year](https://www.odoo.com/documentation/18.0/applications/finance/accounting/reporting/year_end.html#current-year-s-earnings).
This helps users to understand what is happening with this miscellaneous entry using account 999999.

I also propose to add myself as maintainer of this module :)


TDu (approved):
I don't know mis builder, but it is great to see a module revived from the past (did not see the migration to 16.0)

It is strange to see all field names in french and it is something that could be improved ?
But non blocking

l10n_ch_mis_reports/data/mis_report_kpi_bs.xml
```diff
@@ -341,14 +341,43 @@
...
+        <field name="sequence">30</field>
+        <field name="report_id" ref="mis_report_bs" />
+    </record>
+
+    <record id="previous_years_unallocated" model="mis.report.kpi">
+        <field name="name">previous_unallocated</field>
+        <field name="description">Résultat non distribué des années précédentes</field>
+        <field
+            name="expression"
+        >-balu[] -bale[('account_type', 'like', 'equity_unaffected')] +balp[('account_type', 'like', 'equity_unaffected')]</field>
```
divad1196, line 369: (non blocking)
If I understand the module correclty

`previous_unallocated = balp (balance previous) - other stuff`

Would it be more readable to put the `balp` part first?

l10n_ch_mis_reports/data/mis_report_kpi_bs.xml
```diff
@@ -341,14 +341,43 @@
...
+        <field name="report_id" ref="mis_report_bs" />
+    </record>
+
+    <record id="current_year_unallocated" model="mis.report.kpi">
+        <field name="name">current_unallocated</field>
+        <field name="description">Résultat non distribué de l'année en cours</field>
         <field
             name="expression"
-        >-bale[][('account_id.include_initial_balance', '=', False)] -balu[][('account_id.include_initial_balance', '=', False)] - bale[('account_type', '=', 'equity_unaffected')] - balu[('account_type', '=', 'equity_unaffected')]</field>
+        >mis_subreport_pl.CH_R -balp[('account_type', 'like', 'equity_unaffected')]</field>
```
divad1196, line 358: (non blocking)
I don't see `p_l` used elsewhere, maybe you wanted to use it here instead of `mis_subreport_pl.CH_R` ?

divad1196 (approved):
It's mainly a re-arrangement:
 `CH_297` is replaced by `unallocated` and moved from `CH_28` to `CH_2`. 
`unallocated` is now the combination of `current_unallocated` and `previous_unallocated`.
It then needs to compute both `_unallocated` fields

I don't understand much accounting, but the computation seems to make sens.

jguenat:
> I don't know mis builder, but it is great to see a module revived from the past (did not see the migration to 16.0)
> 
> It is strange to see all field names in french and it is something that could be improved ? But non blocking

Thanks for the review !

I'agree it would be better with english. I'll take a look at it :)

l10n_ch_mis_reports/data/mis_report_kpi_bs.xml
jguenat: The "Comptes de résultat" p_l KPI at the bottom is only informational in the balance sheet. I put it there because some accounting users really want to see the raw PL at the bottom of the BS (e.g. when it's already distributed in BS accounts). So I'm kind of reluctant to use it in BS formulas.
You could delete that KPI and the BS still works :)

TDu:
> > I don't know mis builder, but it is great to see a module revived from the past (did not see the migration to 16.0)
> > It is strange to see all field names in french and it is something that could be improved ? But non blocking
> 
> Thanks for the review !
> 
> I'agree it would be better with english. I'll take a look at it :)

It could also be done later, if we want to merge this quicker. Best way is with a ROADMAP.md so we don't forget.

l10n_ch_mis_reports/data/mis_report_kpi_bs.xml
jguenat: not exactly, the p is for period not previous.
+balp is there to negate the unallocated earnings of the current year from the -bale which takes all years (ending balance)
It's better to have it at the end IMO

jguenat:
> > > I don't know mis builder, but it is great to see a module revived from the past (did not see the migration to 16.0)
> > > It is strange to see all field names in french and it is something that could be improved ? But non blocking
> > 
> > 
> > Thanks for the review !
> > I'agree it would be better with english. I'll take a look at it :)
> 
> It could also be done later, if we want to merge this quicker. Best way is with a ROADMAP.md so we don't forget.

It's on the readmap.md :)
- Provides a MIS Builder Report template for the Switzerland VAT
  Declaration.
- Rename reports in english as base language to then be translated
- Translate report names in german

PR #782, opened 2026-03-20
[16.0][FIX] ebill_postfinance: ReceiverParty should not contain empty name

florentx:
In case the customer has an Invoice address, the transmitted name was empty.

The other invoicing  mechanism in Switzerland is `QR-Bill`, in the `l10n_ch` standard add-on: it encodes the name of the Commercial Entity, which is always what we want.

To be consistent, the `ebill_postfinance` should encode the same name in the eBill.

This is the fix.


grindtildeath (approved):

TDu (approved):
Thanks.
I have opened https://github.com/OCA/l10n-switzerland/pull/783 to fix the pre commit issue.

TDu:
Rebased to include the mentioned fix

PR #785, opened 2026-03-27
[18.0[FWP] ebill_postfinance: ReceiverParty should contain the name of the…

TDu:
Forward port of https://github.com/OCA/l10n-switzerland/pull/782

PR #784, opened 2026-03-27
[18.0][FWP] ebill_postfinance: server_reason_code is integer

TDu:
Forward port of https://github.com/OCA/l10n-switzerland/pull/781

PR #790, opened 2026-07-31
[19.0][MIG] ebill_postfinance: Migration to 19.0

Ricardoalso:


ebill_postfinance/models/ebill_postfinance_service.py
```diff
@@ -155,6 +155,6 @@ def get_registration_protocol(self, create_date, archive_data=False):
 
     @api.model
     def cron_update_invoices(self):
-        services = self.search([])
+        services = self.search([], limit=None)
```
ivantodorovich, line 158: I'm curious about this change. Isn't this the default limit?

ivantodorovich:
would be nice to add the pending `test-requirements.txt` so we can see if the tests pass

ebill_postfinance/models/ebill_postfinance_service.py
Ricardoalso: It is indeed the default limit. This is just a small hack to bypass a pre-commit step that warns about searching with an empty domain and the risk of memory errors. Here, with the model `ebill.postfinance.service`, we’re kind of safe and should never face memory errors.

ivantodorovich: Alright, thanks! :)

TDu (approved):
LG thanks

ivantodorovich:
All dependencies are merged.
@Ricardoalso can you rebase & drop the test-requirements? 🙏🏻

ivantodorovich:
I think we need to include this too: 
- https://github.com/OCA/l10n-switzerland/pull/785

ivantodorovich:
And also:
- https://github.com/OCA/l10n-switzerland/pull/784

Ricardoalso:
https://github.com/OCA/l10n-switzerland/pull/785
https://github.com/OCA/l10n-switzerland/pull/784
Both commits picked in here and dependency already merged removed from test-requirements :heavy_check_mark:

ivantodorovich:
Thanks!

@TDu , can you merge? 🙏🏻

TDu:
@Ricardoalso This forward port fix is missing https://github.com/OCA/l10n-switzerland/pull/788, I think.  Could you add it, thanks

Ricardoalso:
> @Ricardoalso This forward port fix is missing #788, I think. Could you add it, thanks

Thanks for pointing it out :pray: 

Done :heavy_check_mark:

TDu:
We still need one more positive review to merge

ivantodorovich (approved):
approved

PR #792, opened 2026-09-09
[17.0] Add module l10n_ch_qr_international

grindtildeath:


l10n_ch_qr_international/models/res_partner_bank.py
```diff
@@ -0,0 +1,31 @@
...
+            and debtor_partner.commercial_partner_id.force_swiss_qr_invoice
+        ):
+            # As super function is returning all the errors reasons in a single string
+            #  we need to split that string using what's used to join the reasons.
+            # Then we can check if there's any other error than the one for partners
+            #  not located in Switzerland or Liechtenstein
+            # If we have any other error, return the result from super call instead
+            #  of trying to remove string that was already translated, with a note
+            #  about what this module already allows
+            qr_errors_nbr = len(res.split("\r\n")) - 2
```
TDu: I think it would be easier and less error prone to copy paste and adapt the code of `_get_error_for_ch_qr` ?
Also knowing that in 18 there is a dedicated function to check the debtor validity https://github.com/odoo/odoo/blob/086c927d6ca4b11efe62d13b4fc377e283764d67/addons/l10n_ch/models/res_bank.py#L237

grindtildeath: I had not seen this before, so I went for a copy of the function. Can you please check again?

TDu (approved):
LG, could you fix the pre-commit and squash.

henrybackman (approved):

